PR #671, opened 2020-04-07
Add support for OSC on web sockets

aalmada:


aalmada:
This implementation uses [Pipelines](https://docs.microsoft.com/en-us/dotnet/standard/io/pipelines). The `Pipe` readers and writers are static methods so these can be reused in the implementation of TCP and UDP, clients and servers, for OSC or any other protocol. They can also be extended to handle files and other transport channels.

[System.IO.Pipelines](https://www.nuget.org/packages/System.IO.Pipelines) package requires NuGet client versions 2.12 or higher. Unfortunately, Bonsai still uses NuGet client version 2.4.0.0. **Which means this PR should not be merged yet.**

PR #888, opened 2022-04-27
Texture filtering user control for IplImageVisualizer

RoboDoig:
Allows a user to switch texture filtering modes on the editor window for IplImage

Button added to bottom status strip, toggles between nearest and linear filter modes for the display texture


RoboDoig:
Doesn't work if multiple IplImage windows are open in editor

PR #914, opened 2022-05-29
Ensure framebuffer resizes one dimension at a time

### LINKED ISSUE #913: Access violation exception when resizing image visualizer canvas
A strange behavior has been observed in specific OS / GPU vendor combinations where resizing an OpenGL framebuffer by resizing the container window will cause a hard crash during the call to `SwapBuffers`.

Although the cause is unknown, and seems to be down to specific issues with drivers and/or software configuration, a possible workaround is to scale the control along only one axis at a time, as documented elsewhere: https://github.com/ocornut/imgui/issues/3321

Although not performance critical, and technically not a critical blocker for everyone, it will noticeably and dramatically affect systems who suffer from this specific configuration, so any mitigation measures which do not harm existing behavior will be appreciated.

glopesdev:
Apparently resizing the GL control only horizontally or vertically prevents `SwapBuffers` from crashing, as described in https://github.com/ocornut/imgui/issues/3321. Building on top of this observation, this PR implements dynamic control resizing by changing only a single dimension per frame until the canvas is fully resized to its target size.

Fixes #913

### DIFF
--- a/Bonsai.Vision.Design/VisualizerCanvas.cs
+++ b/Bonsai.Vision.Design/VisualizerCanvas.cs
@@ -15,6 +15,8 @@ namespace Bonsai.Vision.Design
     {
         bool loaded;
         bool disposed;
+        bool resizing;
+        bool resizeWidth;
         static readonly object syncRoot = string.Intern("A1105A50-BBB0-4EC6-B8B2-B5EF38A9CC3E");
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


glopesdev:
For reference, the issue was tracked down to malware, specifically see comment below:
https://github.com/ocornut/imgui/issues/3321#issuecomment-1146596947

PR #917, opened 2022-06-03
Generalize layout settings and mashup API for arbitrary visualizer composition

### LINKED ISSUE #794: Generalize API for persisting settings of stacked mashup visualizers
There are two main scenarios we are interested in covering here:
- Enable stacked mashup visualizer settings to persist
- Allow hierarchical composition of generic visualizers using "layout" mashups

`TableLayoutPanel` provided a first successful example of this, but the serialization/deserialization API needs to be generalised and moved to the `Design` package.

Serialization format should be kept as uniform as possible. Specifically, keep visualizer qualified type names specified in the layout file, even for stacked settings, and resolve them using the editor context on initialisation.

One challenge is to decide how to keep references to the source nodes that feed into the mashup. There are two options: either keep them at the top level with graph indices as now, or move references to the individual mashup settings.

Either way we probably prefer infrastructure serialization support and avoid propagating specific serialization logic to visualizer implementation code.

### LINKED ISSUE #938: Persist the settings of stacked trajectory traces
Creating image and point sequence mashups allows displaying of visualizer traces with support for specifying memory length in frames. Currently it is not possible to persist these settings, but piggybacking on top of #917 it would be useful to improve the point sequence mashup experience with this functionality.

### LINKED ISSUE #948: Visualizer elements do not propagate correctly for nodes with build dependencies
A visualizer element is defined as the `InspectBuilder` instance representing a specific workflow operator when debugging through the editor (i.e. when double-clicking the node). In most situations, this is simply the `InspectBuilder` which directly wraps the operator.

However, when working with build dependencies, group workflows or nested visualizer operators, the visualizer element is instead retrieved from the closest previous source (e.g. the output of the inner workflow, the source of the property mapping operator, etc). This is to avoid losing access to operator-specific visualizers when indirectly exposing specific node outputs to the top-level of a group, and to allow composition of visualizers at higher-levels, e.g. when using the `VisualizerSink` operator or mashups.

This functionality depends on correctly identifying which operators are "no-op" w.r.t. visualizers in the expression graph. It seems like the current rules do not take into account the case where an operator has build dependencies, i.e. property mapping nodes. This is because there is code generation of calls to ensure correct handling of subscriptions, etc, which is not considered in the detection algorithm inside `InspectBuilder`.

This is a subtle and easy to miss regression, so we should ensure we include a unit test to avoid silent future breaking changes.

glopesdev:
This PR introduces a fairly large code refactoring to unify what we previously called "mashup visualizers" and the regular type visualizers. In short, there is no longer a separation between these two concepts: all visualizers are simply type visualizers.

It also does away with the confusing terminology which existed between type visualizers that combine multiple sources (these now inherit from `MashupVisualizer`) and those visualizers which are designed to overlay some extra visualization on top of an existing base visualizer (these were previously also called "mashup visualizers", but they are now regular type visualizers, for clarity we introduce the suffix `Overlay` to refer to them).

Several bugs were fixed, and backwards compatibility features introduced, to minimize damage to existing layout visualizers.

Fixes #794
Fixes #948
Fixes #938


glopesdev:
For consistency, it would be great if visualizer mashups could go together with the `Visualizer` operator and `WorkflowOutput` (i.e. propagate mashup configurations into the top-level node).

glopesdev:
It looks like a few issues might still be present when reordering, or re-organizing visualizer mappings (e.g. with cut-paste):

image

glopesdev:
Given the flexibility of the current visualizer mapping strategy, it feels like we should take the opportunity to allow specifying which visualizer should be mapped to the mashup. There are a few options:
1. Use the visualizer type which is currently selected on the VisualizerMapping node.
2. Same as above, but explicitly store the type in the VisualizerMapping node properties, so the selection can be made independent of the layout file.
3. Explicitly instantiate the visualizer type in the workflow. This would imply listing all available visualizers in the toolbox, and creating a new general combinator that is able to propagate that visualizer type downstream.

I feel Both 2. and 3. would best approximate the goal of allowing a fully explicit instantiation of the visualizer mappings in the workflow. Option 2. would require only a minimal amount of work to the current core and editor, since it would be simply adding a property to the VisualizerMapping class.

Option 3. would require a set of classes and editor support, and might introduce new unknown problems down the road of implementation regarding the association between user-selected visualizers and a visualizer type explicit set in the workflow (and its settings).

PR #921, opened 2022-06-07
Add support for context-dependent renaming of subjects

### LINKED ISSUE #702: Add support for refactoring subject names
As workflows grow in size and complexity, it becomes more important to provide editor support for refactoring. The first low-hanging fruit is renaming subjects. Subjects are inherently linked by scope so that when the name of the declaration changes, it would be easy to find all the matching subscribe subjects and rename them too.

This could be exposed as a modal UI type editor of the subject name property. Another option is to use the entries in the toolbox itself to identify targets for rename using the context menu.

### LINKED ISSUE #933: Add editor support for finding subject definitions
In complex workflows it can be hard to understand where subjects are declared. We could re-use the "Go to definition" menu option on top of `SubscribeSubject` or `MulticastSubject` instances to allow the editor to find and move the focus to the definition for that subject.

glopesdev:
This PR adds support for context-dependent renaming of subjects across a workflow. The renaming itself happens through the matching subject node in the toolbox, either by right-clicking on the node, or pressing F2 in either the node or the search bar (supporting search and rename).

The operation will correctly identify the source of the subject declaration, and propagate the rename to all dependents and can be initiated either on the original declaration, or in any dependent (subscribe or multicast). Views are refreshed immediately upon rename and undo/redo is supported.

Fixes #702
Fixes #933


glopesdev:
Renaming should not go inside IncludeWorkflow nodes as they are read-only.

glopesdev:
This could also be a good opportunity to address #933

glopesdev:
It would be great to add a "Go to Definition" option also to the context menu for Subjects in the toolbox, so it becomes possible to quickly find the definitions of all listed subjects.

Furthermore, also for each subject listed in the toolbox, it would be great to list as auxiliary information the type of subject in parentheses.

PR #927, opened 2022-06-09
Allow specifying limits for rolling graph builder

### LINKED ISSUE #858: Visualizer graph builders should clear when sequence completes
It is currently clunky or impossible to clear dynamic visualizer graph builders since they do not implement the `SequenceCompleted` method semantics of clearing the graph visualizer (the only possible workaround is to inject empty arrays or structures in the sequence).

This is interesting to include as it allows for the possibility to reset a graph given a specific event.

### LINKED ISSUE #926: Add support for specifying range limits on design visualizers
The current implementation of visualizer builder operators for rolling, line and bar graphs does not provide a status bar for configuring ranges, capacity, etc. This is a well-developed feature in existing visualizers, so it would be useful to have the ability to reuse existing controls where possible.

glopesdev:
This PR adds extended visualization configuration options for the builders in the design visualizers package, to allow specifying range and capacity limits. For consistency, we are also obsoleting redundant visualization properties in the builder nodes.

Fixes #926
Fixes #858

### DIFF
--- a/Bonsai.Design.Visualizers/BarGraph.cs
+++ b/Bonsai.Design.Visualizers/BarGraph.cs
@@ -1,4 +1,5 @@
-﻿using System.Drawing;
+﻿using System.Collections.Generic;
+using System.Drawing;
 using ZedGraph;
 
 namespace Bonsai.Design.Visualizers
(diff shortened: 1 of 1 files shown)


glopesdev:
For certain applications, it is useful to be able to specify "hard" limits for the visualizer, in cases where the range is fixed and known. To allow for both situations, we could expose range properties in the builder classes that take precedence over visualizer editable ranges, but only if they are specified.

glopesdev:
Since `BarGraphVisualizer` dynamically updates the last index value, it would make sense to use it as a dictionary, and consider allowing updating values of pre-existing indexes.

PR #959, opened 2022-07-03
Add operators for reading and writing binary data from a serial port

### LINKED ISSUE #952: Add support for streaming byte buffers into a SerialPort
In order to open up the full potential of the SerialPort interface, it would be great to generalize the functionality of the `SerialStringWrite` node into a full `SerialPortWrite` operator that can take all the possible overloads for serial communication write, `string`, `byte[]`, `ArraySegment<byte>`.

This would greatly expand the potential for implementing custom binary communication protocols over serial port.

glopesdev:
This PR introduces two new operators `SerialRead` and `SerialWrite` which can be used to interface with a serial port using low-level binary representations. The main difference from `SerialStringRead` is that no message terminator is assumed.

Instead, the maximum number of bytes to read is specified, and the sequence will either terminate after this number of bytes is read, or the serial port read timeout elapses. If a source sequence is provided to `SerialRead`, the operator will try to read a byte buffer for each incoming notification.

Conversely, the `SerialWrite` operator will write each of the binary buffers in the source sequence to the serial port. An overload supporting `ArraySegment` is provided.

In preparation for a larger reorganization of the IO package, this PR also refactors the core serial port operators into a new namespace `Bonsai.IO.Ports` mirroring the .NET namespace `System.IO.Ports`. The legacy character based operators `SerialStringRead` and `SerialStringWrite` are also renamed in this new namespace to `SerialReadLine` and `SerialWriteLine`, respectively.

Note that the new `SerialWriteLine` no longer guarantees exclusive access to the port. In general the language is now much better equipped with synchronization primitives using subjects that in general should be used instead of scattering low-level hardware details inside the workflow.

A compatibility layer is provided to ensure all previous workflows can operate as normal (including preserving exclusive access in `SerialStringWrite`). Deprecation notices have been placed on all legacy types.

Fixes #952

### DIFF
--- a/Bonsai.Arduino/ArduinoConfiguration.cs
+++ b/Bonsai.Arduino/ArduinoConfiguration.cs
@@ -1,5 +1,5 @@
 ﻿using System.ComponentModel;
-using Bonsai.IO;
+using Bonsai.IO.Ports;
 
 namespace Bonsai.Arduino
 {
--- a/Bonsai.Arduino/CreateArduino.cs
+++ b/Bonsai.Arduino/CreateArduino.cs
@@ -1,8 +1,8 @@
-﻿using Bonsai.IO;
-using System;
+﻿using System;
 using System.ComponentModel;
 using System.Linq;
 using System.Reactive.Linq;
+using Bonsai.IO.Ports;
 
 namespace Bonsai.Arduino
 {
--- a/Bonsai.System/IO/Ports/BaudRateConverter.cs
+++ b/Bonsai.System/IO/Ports/BaudRateConverter.cs
@@ -1,6 +1,6 @@
 ﻿using System.ComponentModel;
 
-namespace Bonsai.IO
+namespace Bonsai.IO.Ports
 {
     /// <summary>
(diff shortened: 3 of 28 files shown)


glopesdev:
One issue with the current PR is that handling of binary data protocols is usually done in read/write pairs. Having only the `Write` operator will be insufficient in most cases. Unfortunately, the case for the `Read` operator is even more problematic than in character-based protocols with defined encodings.

The main problem is it will be in general hard to define what is a good value for "terminator" characters without having encoding guarantees that the terminator will not be found in the binary encoding of regular numeric values. This means that simply relying on serial big-endian or little-endian encodings without a properly defined protocol is very likely to lead to loss of data if not planned thoroughly.

It might be better to introduce single-byte `Read` operators over the serial port that could be chained in state-machine style to build a sequential reactive parser, simply using fixed-size byte buffers. In many cases, however, one might be better served simply by writing the parser directly in C#.

glopesdev:
The PR has been rewritten to include a source operator that reads a fixed number of bytes from the serial port. The operator can be chained in a sequence or used standalone.

PR #1029, opened 2022-09-07
Update reactive extensions to version 5.0.0

### LINKED ISSUE #1028: Update reactive extensions version
The `Rx-Linq`, `Rx-Main`, etc collection of packages has long been superseded by the modern `System.Reactive` package. Updating the core of Rx has been long overdue and should be done as soon as possible to ease the transition to the modern .NET ecosystem.

glopesdev:
This PR updates the core Reactive Extensions dependency to the latest version. Unfortunately, even though the API is exactly the same, this update breaks binary compatibility with dependent packages and extensions since `System.Reactive` assembly public key token has changed from v2.2.5 to v3.0.

Backwards compatibility is attempted by providing the legacy versions of Rx assemblies for dynamic targeting. This solves most straight-forward bindings by running Rx v2.2.5 and Rx v5.0.0 side-by-side. Unfortunately, indirect method calls to reactive combinators in Bonsai.Core will not survive retargeting of method signatures. For example, all packages using `PublishReconnectable` will require rebuilding to target the new version.

Fixes #1028

### DIFF
--- a/Bonsai.Configuration/ScriptExtensions.cs
+++ b/Bonsai.Configuration/ScriptExtensions.cs
@@ -105,7 +105,7 @@ namespace Bonsai.Configuration
             yield return "System.dll";
             yield return "System.Core.dll";
             yield return "System.Drawing.dll";
-            yield return "System.Reactive.Linq.dll";
+            yield return "System.Reactive.dll";
             yield return "System.Xml.dll";
             yield return "Bonsai.Core.dll";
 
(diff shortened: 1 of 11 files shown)


glopesdev:
A more backwards-compatible path would be to convert legacy packages to target both `net472` and `net6` / `netstandard2.0` and update Rx version only in the latter case. This will allow keeping both .NET Framework and .NET Core bootstrapper and editor applications to coexist side-by-side.

Rx is also currently undergoing a large revision of their codebase in preparation for the release of .NET 8. It looks like the architecture will more closely follow the original division of keeping scheduler assemblies separate from core, so it is probably best to wait for that process to conclude.

PR #1047, opened 2022-09-19
Use new SvgGraphics overload 'string WriteSVGString(SizeF bounds)' to…

monoman:
… simplify and make more resilient the ExportSvg method

### DIFF
--- a/Bonsai.Editor/ExportHelper.cs
+++ b/Bonsai.Editor/ExportHelper.cs
@@ -42,12 +42,7 @@ namespace Bonsai.Editor
             var bounds = graphView.GetLayoutSize();
             var graphics = new SvgNet.SvgGraphics();
             graphView.DrawGraphics(graphics);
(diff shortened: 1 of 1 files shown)


glopesdev:
Thanks, I had noticed the addition of the new overload in SvgNet and of course it makes much more sense to use it instead of manually replacing strings.

The only thing I would change about the PR is the commit message, to make it fit in 50 characters so it is more easy to read in git blame, which we use a lot for debugging and regression testing. Maybe something like "Set SVG bounds without string manipulation".

We are also in the process of finalizing our contribution guidelines so we can confidently start accepting external contributions. This might take a bit longer to resolve as we are currently in the process of testing and preparing a large release containing reorganizations to the editor and core. If you don't mind leaving the PR hanging for a bit, we will merge it ASAP.

monoman:
Sure, take your time...

Rafael Teixeira
O..:.)oooo

PR #1072, opened 2022-10-04
Reuse top-level prefix for all extension base types with explicit XML type declarations

### LINKED ISSUE #1056: XmlSerializer namespace scoping does not include base types
Serializing a combinator with properties coming from a base type in a different namespace, that is itself not part of the workflow (i.e. the base type node is not in the workflow) will generate mangled XML with namespace declarations inline with the property names.

This is also related to the problem of base types "inheriting" the namespace of the derived type without an explicit `XmlTypeAttribute` declaration. Ideally, the workflow XML serializer should walk through the whole inheritance tree of each type and add the correct XML type attribute override to the XML namespace scopes.

glopesdev:
`WorkflowBuilder` serialization procedure has been modified to walk up the inheritance tree of each extension type to search for any base types with explicit `XmlTypeAttribute` declarations. If any exist, they are explicitly added to the set of serializable extension types.

This prevents overly verbose inline XML namespace declarations for properties provided by base types which are not directly used elsewhere in the workflow.

Fixes #1056

### DIFF
--- a/Bonsai.Core.Tests/Constants.cs
+++ b/Bonsai.Core.Tests/Constants.cs
@@ -1,0 +1,7 @@
+﻿namespace Bonsai.Core.Tests
+{
+    internal static class Constants
+    {
+        public const string XmlNamespace = "clr-namespace:Bonsai.Core.Tests;assembly=Bonsai.Core.Tests";
+    }
+}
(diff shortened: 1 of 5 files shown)


glopesdev:
Interestingly the regression test fails locally, but passes in CI. The fear is this issue might actually be framework-version dependent, and therefore attempts to fix it might introduce unspecified behavior.

PR #1204, opened 2022-12-22
Add operators to select a file, or folder, to the `System` package

### LINKED ISSUE #614: Consider adding a SelectFile and SelectFolder source to the system package
**[Original report](https://bitbucket.org/horizongir/bonsai/issue/614) by me.**

----------------------------------------

This would allow an easy user interface for selecting file and folder names for manipulation. The source should simply return the selected string, or pull it out on a trigger.

bruno-f-cruz:
This PR attempts to close #614.

- Both operators allow the user to select a target file, or folder, from a corresponding UI Editor. 

- Both operators have a default `Defer` reactive logic, but can be optionally overloaded with any input sequence, triggering an output event with the currently set `Path` property value.

### DIFF
--- a/Bonsai.System/IO/SelectFile.cs
+++ b/Bonsai.System/IO/SelectFile.cs
@@ -1,0 +1,47 @@
+﻿using Bonsai;
+using System;
+using System.ComponentModel;
+using System.Reactive.Linq;
+
+namespace Bonsai.IO
+{
+    /// <summary>
+    /// An operator that generates a sequence of strings representing
+    /// the path to a user-selected file.
+    /// </summary>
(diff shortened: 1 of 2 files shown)


glopesdev:
In many cases this can be effectively achieved by creating a `PropertySource` from an existing node property with a file or folder picker dialog.

The biggest case I think for having dedicated operators now would be to allow configuration of some of the aspects of the dialog such as available file extension filters, and whether the dialog is open or save dialog, etc. This would imply having more properties on the node and deciding how they would interplay with the value editor (i.e. make sure that externalizing the value property in the end can work with the selected properties of the node).

Issue #1231, opened 2023-01-18
Add operator to query os environment variables


bruno-f-cruz:
Querying environment variables becomes a handy feature, especially when considering the push to make Bonsai easily deployable across different machines seamlessly. 
Some of these variables can effectively make a workflow machine agnostic regarding OS paths, machine name, etc...

This query could be added by simply exposing the [`Environment.GetEnvironmentVariable`](https://learn.microsoft.com/en-us/dotnet/api/system.environment.getenvironmentvariable?view=net-7.0) method already available in .NET

PR #1232, opened 2023-01-18
Add operator to query an environment variable

bruno-f-cruz:
This PR attempts to close https://github.com/bonsai-rx/bonsai/issues/1231.

-  Adds an operator to query an environment variable to `Bonsai.System.IO`.

### DIFF
--- a/Bonsai.System/IO/GetEnvironmentVariable.cs
+++ b/Bonsai.System/IO/GetEnvironmentVariable.cs
@@ -1,0 +1,46 @@
+﻿using Bonsai;
+using System;
+using System.ComponentModel;
+using System.Linq;
+using System.Reactive.Linq;
+
+namespace Bonsai.IO
+{ 
+    /// <summary>
+    /// Represents an operator that returns the value of an OS environment variable.
+    /// </summary>
+    /// 
+    [Description("Returns the value of an environment variable.")]
+    public class GetEnvironmentVariable : Source<string>
+    {
+        [Description("The name of the environment variable to query the value of.")]
+        public string Variable { get; set; }
+
+        /// <summary>
+        /// Generates an observable sequence containing a string with the value
+        /// of the queried environment variable.
+        /// </summary>
+        /// <returns>
+        /// An observable sequence containing the value of the queried
+        /// environment variable.
+        /// </returns>
+        public override IObservable<string> Generate()
+        {
+            return Observable.Return(Environment.GetEnvironmentVariable(Variable));
+        }
+
+        /// <summary>
+        /// Generates an observable sequence containing a string with the value
+        /// of the queried environment variable, triggered with any input
+        /// sequence
+        /// </summary>
+        /// <returns>
+        /// An observable sequence containing the value of the queried
+        /// environment variable.
+        /// </returns>
+        public IObservable<string> Generate<TSource>(IObservable<TSource> source)
+        {
+            return source.Select(x => { return Environment.GetEnvironmentVariable(Variable); });
+        }
+    }
+}


Bonsai.System/IO/GetEnvironmentVariable.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.ComponentModel;
+using System.Linq;
+using System.Reactive.Linq;
+
+namespace Bonsai.IO
+{ 
+    /// <summary>
+    /// Represents an operator that returns the value of an OS environment variable.
+    /// </summary>
+    ///
```
glopesdev, line 12: Remove empty line

Bonsai.System/IO/GetEnvironmentVariable.cs
```diff
@@ -0,0 +1,46 @@
+﻿using Bonsai;
+using System;
+using System.ComponentModel;
+using System.Linq;
+using System.Reactive.Linq;
+
+namespace Bonsai.IO
+{ 
+    /// <summary>
+    /// Represents an operator that returns the value of an OS environment variable.
```
glopesdev, line 10: More generally, environment variables are specified per-process, so they might be different than OS variables. I would probably try to keep the wording of the original docs and say something like:

```c#
/// Represents an operator that gets the value of an environment variable for the current process.
```

Bonsai.System/IO/GetEnvironmentVariable.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// </summary>
+    /// 
+    [Description("Returns the value of an environment variable.")]
```
glopesdev, line 13: This should be consistent with whatever is the final wording, for example:

```c#
[Description("Gets the value of an environment variable for the current process.")]
```

Bonsai.System/IO/GetEnvironmentVariable.cs
```diff
@@ -0,0 +1,46 @@
...
+    public class GetEnvironmentVariable : Source<string>
+    {
+        [Description("The name of the environment variable to query the value of.")]
```
glopesdev, line 16: The property needs doc strings as well, for example:

```c#
/// <summary>
/// Gets or sets the name of the environment variable to query the value of.
/// </summary>
```

Bonsai.System/IO/GetEnvironmentVariable.cs
```diff
@@ -0,0 +1,46 @@
...
+        public string Variable { get; set; }
```
glopesdev, line 17: This name is consistent with the parameter name of the `GetEnvironmentVariable` method, but I wonder if it would make more sense to just use `Name` for this so we can take advantage of the option to group editing and externalizing of similar properties in the editor.

Bonsai.System/IO/GetEnvironmentVariable.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// Generates an observable sequence containing a string with the value
+        /// of the queried environment variable.
+        /// </summary>
+        /// <returns>
+        /// An observable sequence containing the value of the queried
+        /// environment variable.
+        /// </returns>
```
glopesdev, line 26: For consistency with documentation of similar operators, I would use something like:

```c#
        /// <summary>
        /// Gets the value of the specified environment variable for the current process
        /// and surfaces it through an observable sequence.
        /// </summary>
        /// <returns>
        /// A sequence containing the value of the specified environment variable. The
        /// value will be <see langword="null"/> if the environment variable is not found.
        /// </returns>
```

Bonsai.System/IO/GetEnvironmentVariable.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// of the queried environment variable, triggered with any input
+        /// sequence
+        /// </summary>
+        /// <returns>
+        /// An observable sequence containing the value of the queried
+        /// environment variable.
+        /// </returns>
+        public IObservable<string> Generate<TSource>(IObservable<TSource> source)
+        {
+            return source.Select(x => { return Environment.GetEnvironmentVariable(Variable); });
```
glopesdev, line 43: For conciseness this can be written as:

```c#
return source.Select(_ => Environment.GetEnvironmentVariable(Variable));
```

Bonsai.System/IO/GetEnvironmentVariable.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// Generates an observable sequence containing a string with the value
+        /// of the queried environment variable, triggered with any input
+        /// sequence
+        /// </summary>
+        /// <returns>
+        /// An observable sequence containing the value of the queried
+        /// environment variable.
+        /// </returns>
```
glopesdev, line 40: Similar to the above, small edits:

```c#
        /// <summary>
        /// Gets the value of the specified environment variable for the current process
        /// whenever an observable sequence emits a notification.
        /// </summary>
        /// <typeparam name="TSource">
        /// The type of the elements in the <paramref name="source"/> sequence.
        /// </typeparam>
        /// <param name="source">
        /// The sequence of notifications used to get the value of the environment variable.
        /// </param>
        /// <returns>
        /// A sequence containing the current values of the specified environment variable.
        /// The value may be <see langword="null"/> if the environment variable is not found.
        /// </returns>
```

Bonsai.System/IO/GetEnvironmentVariable.cs
```diff
@@ -0,0 +1,46 @@
+﻿using Bonsai;
```
glopesdev, line 1: This using should not be needed as the operator is declared inside the `Bonsai` namespace.

Bonsai.System/IO/GetEnvironmentVariable.cs
```diff
@@ -0,0 +1,46 @@
...
+using System;
+using System.ComponentModel;
+using System.Linq;
```
glopesdev, line 4: This using `System.Linq` may not be needed.

glopesdev (changes requested):

Bonsai.System/IO/GetEnvironmentVariable.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// <param name="source">
+        /// The sequence of notifications used to get the value of the environment variable.
+        /// </param>
+        /// <returns>
+        /// A sequence containing the current values of the specified environment variable.
+        /// The value may be <see langword="null"/> if the environment variable is not found.
+        /// </returns>
+        public IObservable<string> Generate<TSource>(IObservable<TSource> source)
+        {
+            return source.Select(_ => { return Environment.GetEnvironmentVariable(Name); });
```
glopesdev, line 49: Is there a reason why it is not possible to use the more compact lambda notation below? I haven't tried it myself but it should work.

```c#
return source.Select(_ => Environment.GetEnvironmentVariable(Name));
```

glopesdev (changes requested):
All sounds great, just a minor edit pending.

glopesdev (approved):

Bonsai.System/IO/GetEnvironmentVariable.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using System.ComponentModel;
+using System.Reactive.Linq;
+
+namespace Bonsai.IO
+{
+    /// <summary>
+    /// Represents an operator that gets the value of an environment variable for the current process.
+    /// </summary>
+    ///
```
glopesdev, line 10: Remove empty line?

glopesdev (approved):

Issue #1727, opened 2024-04-06
Allow script extensions to declare unsafe blocks


glopesdev:
The current compiler parameters for custom extensions include only `/optimize`.

Given the increase in low-level bit-manipulation code extensions, it would be very useful if we could also add `/unsafe` whenever the `Extensions.csproj` file contains the declaration `<AllowUnsafeBlocks>true</AllowUnsafeBlocks>`.

Issue #1740, opened 2024-04-15
Workflows running from the command line do not terminate the process on completion


glopesdev:
This was a regression likely introduced on release of 2.8.0 with the change to use inter-process communication instead of AppDomains for isolating workflow execution.

To reproduce:
```
bonsai.exe --no-editor emptyWorkflow.bonsai
```

Issue #778, opened 2021-04-23
Add support for recent language constructs in extension scripts


glopesdev:
The current compiler used by the extension script provider uses the C# language services provided by CodeDOM, which is limited to only earlier language features.

To extend support to more modern constructs, the Roslyn compiler libraries should be used to build extension scripts and achieve parity between library development in VS and free-form scripting.

PR #1753, opened 2024-04-22
Allow script extensions to declare unsafe blocks

PathogenDavid:
The setting can be toggled by modifying `Extensions.csproj`, same as any other C# project.

This change also improves the robustness of the `UseWindowsForms` check since I modified it to use the helper I added for this change.

In particular it now only checks in appropriate locations within the project file, and perhaps more importantly it disregards the case of `true` vs `True`. At some point\* Visual Studio started using the latter when these settings are changed via the UI, so anyone using it would've had `True` treated as false.

<sub>(\*I assume so at least, I typically just modify my `csproj` by hand and I don't have an old version of Visual Studio to check at the moment–I would guess it changed when they redid the project configuration tool.)</sub>

Opening the whole project file to check a single value makes me slightly itchy, but it's consistent with similar logic in `ScriptExtensions` and I didn't want to get lost in the weeds refactoring.

Fixes https://github.com/bonsai-rx/bonsai/issues/1727

### DIFF
--- a/Bonsai.Configuration/ScriptExtensions.cs
+++ b/Bonsai.Configuration/ScriptExtensions.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Reflection;
 using System.Xml;
 using System.Xml.Linq;
+using System.Xml.XPath;
 
 namespace Bonsai.Configuration
 {
@@ -20,6 +21,7 @@ namespace Bonsai.Configuration
         const string PackageVersionAttribute = "Version";
         const string UseWindowsFormsElement = "UseWindowsForms";
         const string ItemGroupElement = "ItemGroup";
+        const string AllowUnsafeBlocksElement = "AllowUnsafeBlocks";
         const string ProjectFileTemplate = @"<Project Sdk=""Microsoft.NET.Sdk"">
 
   <PropertyGroup>
@@ -98,6 +100,12 @@ namespace Bonsai.Configuration
             };
         }
 
+        static XElement GetProperty(XDocument document, string key)
+            => document.XPathSelectElement($"/Project/PropertyGroup/{key}");
+
+        static bool? GetBoolProperty(XDocument document, string key)
+            => GetProperty(document, key)?.Value?.Equals("true", StringComparison.InvariantCultureIgnoreCase);
+
         public IEnumerable<string> GetAssemblyReferences()
         {
             yield return "System.dll";
@@ -114,8 +122,7 @@ namespace Bonsai.Configuration
             if (!File.Exists(ProjectFileName)) yield break;
             using var stream = File.OpenRead(ProjectFileName);
             var document = LoadProjectDocument(stream);
-            var useWindowsForms = document.Descendants(XName.Get(UseWindowsFormsElement)).FirstOrDefault();
-            if (useWindowsForms != null && useWindowsForms.Value == "true")
+            if (GetBoolProperty(document, UseWindowsFormsElement) ?? false)
             {
                 yield return "System.Windows.Forms.dll";
             }
@@ -193,6 +200,13 @@ namespace Bonsai.Configuration
             }
 
             File.WriteAllText(ProjectFileName, root.ToString(SaveOptions.DisableFormatting));
+        }
+
+        public bool GetAllowUnsafeBlocks()
+        {
+            using var stream = File.OpenRead(ProjectFileName);
+            var document = LoadProjectDocument(stream);
+            return GetBoolProperty(document, AllowUnsafeBlocksElement) ?? false;
         }
 
         public void Dispose()
--- a/Bonsai.Configuration/ScriptExtensions.cs
+++ b/Bonsai.Configuration/ScriptExtensions.cs
@@ -1,5 +1,4 @@
 ﻿using NuGet.Configuration;
-using NuGet.Versioning;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -7,22 +6,13 @@ using System.Linq;
 using System.Reflection;
 using System.Xml;
 using System.Xml.Linq;
-using static Bonsai.Configuration.ScriptExtensionsProjectMetadata;
 
 namespace Bonsai.Configuration
 {
     public class ScriptExtensions : IDisposable
     {
-        const string OuterIndent = "  ";
-        const string InnerIndent = "    ";
         const string DefaultProjectFileName = "Extensions.csproj";
-        const string ProjectFileTemplate = @"<Project Sdk=""Microsoft.NET.Sdk"">
 
-  <PropertyGroup>
-    <TargetFramework>net472</TargetFramework>
-  </PropertyGroup>
-
-</Project>";
         readonly IDictionary<string, PackageReference> packageMap;
         readonly PackageConfiguration packageConfiguration;
         readonly TempDirectory assemblyFolder;
@@ -96,67 +86,21 @@ namespace Bonsai.Configuration
             return new ScriptExtensionsProjectMetadata(document);
         }
 
+        public void UpdateProjectMetadata(ScriptExtensionsProjectMetadata projectMetadata)
+            => File.WriteAllText(ProjectFileName, projectMetadata.GetProjectXml());
+
         public void AddAssemblyReferences(IEnumerable<string> assemblyReferences)
         {
-            XElement root;
-            if (!File.Exists(ProjectFileName))
-            {
-                root = XElement.Parse(ProjectFileTemplate, LoadOptions.PreserveWhitespace);
+            var projectMetadata = LoadProjectMetadata();
+            if (!projectMetadata.Exists)
                 EnsureNuGetSettings();
-            }
-            else root = XElement.Load(ProjectFileName, LoadOptions.PreserveWhitespace);
-
-            var projectReferences = root.Descendants(PackageReferenceElement).ToArray();
-            var lastReference = projectReferences.LastOrDefault();
 
             var packageReferences = assemblyReferences
                 .Select(assemblyName => packageConfiguration.GetAssemblyPackageReference(assemblyName, packageMap))
-                .Where(package => package != null);
-            foreach (var reference in packageReferences)
-            {
-                var includeAttribute = new XAttribute(PackageIncludeAttribute, reference.Id);
-                var versionAttribute = new XAttribute(PackageVersionAttribute, reference.Version);
-                var existingReference = (from element in projectReferences
-                                         let id = element.Attribute(PackageIncludeAttribute)
-                                         where id != null && id.Value == reference.Id
-                                         select element).FirstOrDefault();
-                if (existingReference != null)
-                {
-                    var version = existingReference.Attribute(PackageVersionAttribute);
-                    if (version == null) existingReference.Add(versionAttribute);
-                    else if (NuGetVersion.Parse(version.Value) < NuGetVersion.Parse(reference.Version))
-                    {
-                        version.SetValue(reference.Version);
-                    }
+                .Where(package => package is not null);
 
-                    continue;
-                }
-
-                var referenceElement = new XElement(PackageReferenceElement, includeAttribute, versionAttribute);
-                if (lastReference == null)
-                {
-                    var itemGroup = new XElement(ItemGroupElement);
-                    itemGroup.Add(Environment.NewLine + InnerIndent);
-                    itemGroup.Add(referenceElement);
-                    itemGroup.Add(Environment.NewLine + OuterIndent);
-
-                    root.Add(OuterIndent);
-                    root.Add(itemGroup);
-                    root.Add(Environment.NewLine);
-                    root.Add(Environment.NewLine);
-                }
-                else
-                {
-                    lastReference.AddAfterSelf(referenceElement);
-                    if (lastReference.PreviousNode != null && lastReference.PreviousNode.NodeType >= XmlNodeType.Text)
-                    {
-                        lastReference.AddAfterSelf(lastReference.PreviousNode);
-                    }
-                }
-                lastReference = referenceElement;
-            }
-
-            File.WriteAllText(ProjectFileName, root.ToString(SaveOptions.DisableFormatting));
+            projectMetadata = projectMetadata.AddPackageReferences(packageReferences);
+            UpdateProjectMetadata(projectMetadata);
         }
 
         public void Dispose()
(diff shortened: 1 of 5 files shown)


Bonsai.Configuration/ScriptExtensions.cs
```diff
@@ -195,6 +202,13 @@ public void AddAssemblyReferences(IEnumerable<string> assemblyReferences)
             File.WriteAllText(ProjectFileName, root.ToString(SaveOptions.DisableFormatting));
         }
 
+        public bool GetAllowUnsafeBlocks()
```
glopesdev, line 205: I agree it is annoying to have to open the file multiple times just to retrieve the value of these properties, and that indeed the refactoring seems painfully non-trivial.

I guess the way I might do it is have all these methods take an `XDocument` as argument and make `LoadProjectDocument` non-static and public:

```c#
        public XDocument LoadProjectDocument()
        {
            if (!File.Exists(ProjectFileName)) return new XDocument();
            using var stream = File.OpenRead(ProjectFileName);
            using var reader = XmlReader.Create(stream, GetXmlReaderSettings());
            return XDocument.Load(reader, LoadOptions.None);
        }
```

Because the `XDocument` instance is created and loaded into memory all at once we don't have to worry about keeping track of stream dispose calls, so the call sites would just need to use `LoadProjectDocument` first and pass the document into each of the methods to get the different bits of information, e.g. for the `GetAllowUnsafeBlocks` method:

```c#
        public bool GetAllowUnsafeBlocks(XDocument document)
            => GetBoolProperty(document, AllowUnsafeBlocksElement);
```

Incidentally this also optimizes another call to `GetPackageReferences` which was already opening the file twice inside `ScriptExtensionsProvider` and which could now be refactored like so:

```c#
            var projectDocument = scriptEnvironment.LoadProjectDocument();
            using (var cacheContext = new SourceCacheContext())
            {
                var projectReferences = from id in scriptEnvironment.GetPackageReferences(projectDocument)
                                        from assemblyReference in FindAssemblyReferences(
                                            projectFramework,
                                            dependencyResource,
                                            localPackageResource,
                                            cacheContext,
                                            packageId: id)
                                        select assemblyReference;
                assemblyNames.AddRange(scriptEnvironment.GetAssemblyReferences(projectDocument));
                assemblyNames.AddRange(projectReferences);
            }
```

The call site at `DependencyInspector` would have to be refactored as well, but I guess that would be it.

Although some of these are "public" classes, the entire `Bonsai.Configuration` assembly is internalized inside the bootstrapper projects, so never visible to downstream consumers and therefore it shouldn't be an API breaking change anyway.

Let me know what you think. I didn't actually try and test this refactoring, so there might be something important I missed, for example in the assumption that creating an empty `XDocument` works for the case where the project file does not exist.

Bonsai.Configuration/ScriptExtensionsProvider.cs
```diff
@@ -143,15 +143,26 @@ string GetAssemblyLocation(string fileName)
...
+                compilerParameters.CompilerOptions += " /unsafe";
+            }
+            
             if (!includeDebugInformation)
             {
-                compilerParameters.CompilerOptions = "/optimize";
+                compilerParameters.CompilerOptions += " /optimize";
             }
 
+            compilerParameters.CompilerOptions = compilerParameters.CompilerOptions.TrimStart();
```
glopesdev, line 164: I guess an alternative to this would be to have a list of strings and assign only once at the end by concatenating all options using white space as a separator. Might be useful if later we keep adding more compiler options.

glopesdev (changes requested):
@PathogenDavid Looks great, I like the `XPath` solution and I definitely didn't know about the VS shenanigans with `true` vs `True` so thanks for calling that to attention.

I suggested a possible approach to refactor the `XDocument` business, simply because the file was already being opened twice before. I had left it like this before pretty much for the same reasons to avoid a painful refactoring, but I guess third time's the charm as they say.

Bonsai.Configuration/ScriptExtensions.cs
```diff
@@ -114,8 +122,7 @@ public IEnumerable<string> GetAssemblyReferences()
             if (!File.Exists(ProjectFileName)) yield break;
             using var stream = File.OpenRead(ProjectFileName);
             var document = LoadProjectDocument(stream);
-            var useWindowsForms = document.Descendants(XName.Get(UseWindowsFormsElement)).FirstOrDefault();
-            if (useWindowsForms != null && useWindowsForms.Value == "true")
+            if (GetBoolProperty(document, UseWindowsFormsElement) ?? false)
```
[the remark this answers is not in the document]
PathogenDavid: My logic is that I read `GetBoolProperty(...) ?? false` as `GetBoolProperty or default to false if absent` but I'm good either way

Bonsai.Configuration/ScriptExtensions.cs
```diff
@@ -98,6 +100,12 @@ static XmlReaderSettings GetXmlReaderSettings()
             };
         }
 
+        static XElement GetProperty(XDocument document, string key)
+            => document.XPathSelectElement($"/Project/PropertyGroup/{key}");
+
+        static bool? GetBoolProperty(XDocument document, string key)
+            => GetProperty(document, key)?.Value?.Equals("true", StringComparison.InvariantCultureIgnoreCase);
```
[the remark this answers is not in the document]
PathogenDavid: The logic here was to make the caller consider what the default should actually be when the property is absent since it might not be `false`, but it almost always is so that's fine

Bonsai.Configuration/ScriptExtensionsProvider.cs
PathogenDavid: I almost wrote it that way but thought it'd come across as extra noise, plus the string command line options will probably go away with https://github.com/bonsai-rx/bonsai/issues/778 anyway. I'll change it to avoid the itch either way though.

Bonsai.Configuration/ScriptExtensions.cs
PathogenDavid: > Although some of these are "public" classes, the entire `Bonsai.Configuration` assembly is internalized inside the bootstrapper projects, so never visible to downstream consumers and therefore it shouldn't be an API breaking change anyway.

This was an important piece of context I was definitely missing. The class being public is a lot of what made me hesitant to do anything that'd change the API or might cause subtle changes in behavior.

Originally I was also hesitating to refactor things since my gut said this all might be replaceable by the [Roslyn Workspace model](https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/work-with-workspace) when https://github.com/bonsai-rx/bonsai/issues/778 happens, but looking a bit closer I'm noticing that [Bonsai.DependencyInspector.GetPackageDependencies](https://github.com/bonsai-rx/bonsai/blob/9357b9ee76efee403a2717b225acb02307681bf2/Bonsai/DependencyInspector.cs#L118) relies on getting a list of package references and I'm not sure if that's something Roslyn can provide. (IIRC the workspace model is pretty limited as it has a very Roslyn-centric view of projects.)

Anyway, thoughts aside I think the general strategy is good for avoiding any unintended behavior changes I'll test and polish it slightly and push a commit later today.

Bonsai.Configuration/ScriptExtensions.cs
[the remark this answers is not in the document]
PathogenDavid: (Also after letting that thought bounce around in my head, that logic doesn't actually reflect how MSBuild works regardless since it doesn't actually have booleans.)

Bonsai.Configuration/ScriptExtensions.cs
PathogenDavid: Make that tomorrow, https://github.com/bonsai-rx/bonsai/issues/1740 turned into a rabbit hole

Bonsai.Configuration/ScriptExtensions.cs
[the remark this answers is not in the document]
PathogenDavid: Done in refactored version

Bonsai.Configuration/ScriptExtensionsProvider.cs
PathogenDavid: Done

Bonsai.Configuration/ScriptExtensions.cs
PathogenDavid: This is basically what I ended up doing except I wrapped the `XDocument` in a new type `ScriptExtensionsProjectMetadata` which takes all of the read-only project helpers.

There's definitely more elegant refactorings out there but I think it's probably better to keep things simple.

Bonsai.Configuration/ScriptExtensions.cs
```diff
@@ -7,6 +7,7 @@
 using System.Reflection;
 using System.Xml;
 using System.Xml.Linq;
+using static Bonsai.Configuration.ScriptExtensionsProjectMetadata;
```
glopesdev, line 10: @PathogenDavid refactoring looks great, and it made me wonder whether for consistency we should move the `ProjectFileTemplate` and `AddAssemblyReferences` method inside `ScriptExtensionsProjectMetadata`.

This would eliminate the need to make the constants internal and more cleanly separate out all metadata file manipulation.

To make this work cleanly we would probably need to change the signature to use package references instead of assembly references (to keep out the dependency on `PackageConfiguration`), move all file IO outside (including the call to `EnsureNuGetSettings`) and return a new instance of project metadata if we want to keep the struct strictly immutable.

Maybe something like this?

```c#
        public ScriptExtensionsProjectMetadata AddPackageReferences(IEnumerable<PackageReference> packageReferences)
        {
            var root = projectDocument?.Root;
            root ??= XElement.Parse(ProjectFileTemplate, LoadOptions.PreserveWhitespace);

            var projectReferences = root.Descendants(PackageReferenceElement).ToArray();
            var lastReference = projectReferences.LastOrDefault();
            foreach (var reference in packageReferences)
            {
                var includeAttribute = new XAttribute(PackageIncludeAttribute, reference.Id);
                var versionAttribute = new XAttribute(PackageVersionAttribute, reference.Version);
                var existingReference = (from element in projectReferences
                                         let id = element.Attribute(PackageIncludeAttribute)
                                         where id != null && id.Value == reference.Id
                                         select element).FirstOrDefault();
                if (existingReference != null)
                {
                    var version = existingReference.Attribute(PackageVersionAttribute);
                    if (version == null) existingReference.Add(versionAttribute);
                    else if (NuGetVersion.Parse(version.Value) < NuGetVersion.Parse(reference.Version))
                    {
                        version.SetValue(reference.Version);
                    }

                    continue;
                }

                var referenceElement = new XElement(PackageReferenceElement, includeAttribute, versionAttribute);
                if (lastReference == null)
                {
                    var itemGroup = new XElement(ItemGroupElement);
                    itemGroup.Add(Environment.NewLine + InnerIndent);
                    itemGroup.Add(referenceElement);
                    itemGroup.Add(Environment.NewLine + OuterIndent);

                    root.Add(OuterIndent);
                    root.Add(itemGroup);
                    root.Add(Environment.NewLine);
                    root.Add(Environment.NewLine);
                }
                else
                {
                    lastReference.AddAfterSelf(referenceElement);
                    if (lastReference.PreviousNode != null && lastReference.PreviousNode.NodeType >= XmlNodeType.Text)
                    {
                        lastReference.AddAfterSelf(lastReference.PreviousNode);
                    }
                }
                lastReference = referenceElement;
            }

            return new ScriptExtensionsProjectMetadata(new XDocument(root));
        }
```

Then the `AddAssemblyReferences` method itself on `ScriptExtensions` could be reduced to:

```c#
        public void AddAssemblyReferences(IEnumerable<string> assemblyReferences)
        {
            var projectMetadata = LoadProjectMetadata();
            if (!projectMetadata.Exists)
            {
                EnsureNuGetSettings();
            }

            var packageReferences = assemblyReferences
                .Select(assemblyName => packageConfiguration.GetAssemblyPackageReference(assemblyName, packageMap))
                .Where(package => package != null);

            projectMetadata = projectMetadata.AddPackageReferences(packageReferences);
            File.WriteAllText(ProjectFileName, projectMetadata.ToString());
        }
```

This assumes there is also a `ToString` override on `ScriptExtensionsProjectMetadata` to emit the formatted output. Below is a simple but probably error prone implementation:

```c#
        public override string ToString()
        {
            return projectDocument?.Root.ToString(SaveOptions.DisableFormatting);
        }
```

Bonsai.Configuration/ScriptExtensionsProjectMetadata.cs
```diff
@@ -0,0 +1,61 @@
...
+        internal const string AllowUnsafeBlocksElement = "AllowUnsafeBlocks";
+
+        internal ScriptExtensionsProjectMetadata(XDocument projectDocument)
+            => this.projectDocument = projectDocument;
+
+        private XElement GetProperty(string key)
+            => projectDocument?.XPathSelectElement($"/Project/PropertyGroup/{key}");
+
+        private bool GetBoolProperty(string key)
+            => String.Equals(GetProperty(key)?.Value, "true", StringComparison.InvariantCultureIgnoreCase);
```
glopesdev, line 29:
```suggestion
            => string.Equals(GetProperty(key)?.Value, "true", StringComparison.InvariantCultureIgnoreCase);
```

Nitpick, just for consistency with the rest of the codebase.

glopesdev (changes requested):

Bonsai.Configuration/ScriptExtensions.cs
PathogenDavid: > ```csharp
> .Where(package => package is not null);
> ```

I know the old implementation had this, but is this actually a good thing to do? This basically means we couldn't figure out how to reference the requested assembly, should we throw instead?

Bonsai.Configuration/ScriptExtensionsProjectMetadata.cs
PathogenDavid: Coulda sworn I fixed that, thanks! Added the rule to the editorconfig too.

PathogenDavid:
(Moved this out to a comment so it doesn't get lost when I resolve the conversation)

> ```csharp
> .Where(package => package is not null);
> ```

I know the old implementation had this, but is this actually a good thing to do? This basically means we couldn't figure out how to reference the requested assembly, should we throw instead?

Bonsai.Configuration/ScriptExtensions.cs
PathogenDavid: Done!

glopesdev (approved):

PR #1755, opened 2024-04-23
Fix workflows run from the command line not exiting upon completion

PathogenDavid:
(Submitted as a draft since I want to double-check some things tomorrow morning and probably revisit the wording in the explanation comment.)

This issue was actually two different issues in a trench coat:

---------------------

For a completely empty workflow (and perhaps other simple workflows?) the finally action will execute on the main thread, which causes `Application.Exit` to be called before `Application.Run` which means it doesn't exit.

---------------------

The second scenario is a workflow that end naturally and has a visualizer enabled. ([Example workflow](https://github.com/bonsai-rx/bonsai/files/15083579/take10.zip))

In this case the finally action will execute on a background thread. `Application.Exit` invokes `FormClosing` and `FormClosed` event handlers directly. This means those handlers are invoked off the main thread, which upsets *many* `FormClosed` handlers within Bonsai.

As an added treat, any exceptions thrown from finally actions are swallowed unceremoniously. As a result, one of the many unhappy `FormClosed` handlers interrupts `Application.Exit` and the signal to exit the message pump never reaches the main thread.

As a helpful side-effect, this means those exceptions are no longer ignored.

---------------------

I also fixed the NotifyIcon getting leaked (which was the underlying clause of it hanging around when the application exited--which is what I assume the old explicit hide was trying to prevent.)

As an aside, I noticed `Observable.Finally` is used all over within Bonsai to run dispose methods and such. It seems like it'd be a good idea to investigate that further and see if it's expected behavior from Rx.NET or maybe a bug.

Fixes https://github.com/bonsai-rx/bonsai/issues/1740

### DIFF
--- a/Bonsai.Editor/WorkflowRunner.cs
+++ b/Bonsai.Editor/WorkflowRunner.cs
@@ -70,20 +70,28 @@ namespace Bonsai.Editor
             contextMenu.Items.Add(new ToolStripSeparator());
             contextMenu.Items.Add(new ToolStripMenuItem("Stop", null, (sender, e) => cts.Cancel()));
 
-            var notifyIcon = new NotifyIcon();
+            using var notifyIcon = new NotifyIcon();
             notifyIcon.Icon = Properties.Resources.Icon;
             notifyIcon.Text = Path.GetFileName(fileName);
             notifyIcon.ContextMenuStrip = contextMenu;
             notifyIcon.Visible = true;
+
+            using var synchronizationContext = new WindowsFormsSynchronizationContext();
             runtimeWorkflow.Finally(() =>
             {
-                notifyIcon.Visible = false;
-                Application.Exit();
+                // Posting the exit to the main thread's winforms sync context is important for two reasons:
+                // 1) When this finally action executes on the main thread we need to defer exiting until
+                //    Application.Run, otherwise we're trying to exit a message loop which hasn't even started.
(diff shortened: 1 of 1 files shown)


glopesdev:
@PathogenDavid great investigation, both of these are very significant regressions. Indeed when a workflow is executed in the editor all disposal events are marshalled correctly to the main window, but I failed to account for the asynchronous dispose in this case, interesting how this was able to fly under the radar for so long!

I will wait for the full PR, but just two quick comments:
* Do we need to worry about disposing the `synchronizationContext` itself, i.e. should there be a `using` statement there too? Maybe not in the case of the regular application path, but I'm thinking here about the case where someone calls `Run` programmatically.
* Re. the use of `Observable.Finally`, I would of course be very interested to audit the caveats of this pattern. Feel free to open a new issue for this if you find anything.

PathogenDavid:
Good call on disposing the sync context.

I pushed better wording for my comment and that fix.

I got lost in the weeds a bit looking into the `Observable.Finally` thing and ended up filing an issue here: https://github.com/bonsai-rx/bonsai/issues/1756 Maybe since you know Rx.NET better than I do something stands out to you there? It looks like it might become *more* of a problem when we upgrade to Rx.NET 6.0.0, so it might be worth fully investigating for Bonsai 3.0

glopesdev (approved):

PR #1760, opened 2024-04-26
Prevent empty category nodes from being dragged into the workflow

### LINKED ISSUE #1759: When the workflow/subject categories are empty, dragging them to the workflow results in `NullReferenceException`
Noticed while monkeying around trying to reproduce https://github.com/bonsai-rx/bonsai/issues/1591

When the Workflow/Subject toolbox categories are empty (as pictured below), it's possible to drag them to the Workflow. Doing so results in a `NullReferenceException` in [`GetToolboxElementCategory`](https://github.com/bonsai-rx/bonsai/blob/50627731bfbf265a08e06f22ce6e4ec85053dc9d/Bonsai.Editor/GraphView/WorkflowGraphView.cs#L128-L138) due to `elementCategories` being `null`.

image

Considering they stop being draggable once they have members, I assume they aren't supposed to be draggable ever.

glopesdev:
The current check for deciding whether a toolbox node can be dragged into the workflow was based on testing for number of children, i.e. operator nodes are always leaf nodes either in hierarchical or search mode.

However, category nodes can also be leaf nodes if there are no operators of that category. This can easily happen for the `Workflow` and `Subject` categories. An additional restriction to prevent this is to check whether the node actually has a non-null content `Tag`.

Fixes #1759

### DIFF
--- a/Bonsai.Editor/EditorForm.cs
+++ b/Bonsai.Editor/EditorForm.cs
@@ -1541,7 +1541,9 @@ namespace Bonsai.Editor
 
         private void toolboxTreeView_ItemDrag(object sender, ItemDragEventArgs e)
         {
-            if (e.Item is TreeNode selectedNode && selectedNode.GetNodeCount(false) == 0)
+            if (e.Item is TreeNode selectedNode &&
+                selectedNode.Tag is not null &&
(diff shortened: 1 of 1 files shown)


PathogenDavid (approved):
Figured it was probably something like that 😁

Issue #1670, opened 2024-02-03
Consider replacing newlines with empty space in text visualizer


glopesdev:
The new text visualizer now strips line endings with empty strings so each value fits in a single row of the buffer. An easy way to preserve some extra readability would be to replace the line endings with a single white space.

glopesdev:
Another option would be to replace all hidden control characters with a rendition of the respective control character, e.g. https://en.wikipedia.org/wiki/Unicode_control_characters#Control_pictures

PathogenDavid:
I did a partial implementation of using control picture replacements. While it does work, legibility with the default fonts on Windows 10 is not ideal:

![image](https://github.com/bonsai-rx/bonsai/assets/278957/cf698bfd-2ed2-4e0f-a491-24c498dfd7d3)

Not sure if it maybe looks better on Windows 11. (During dev club I noticed the control pictures did look different on your screen.) If you want to check here's [my WIP branch](https://github.com/PathogenDavid/bonsai/tree/5108a7c307f7db476c058ef6ca3b872df71e80cb) and [my test file](https://github.com/bonsai-rx/bonsai/files/15069199/test.txt).


I did notice that this visualizer uses `RichTextBox`, so in theory we could dim the glyphs which would probably help. That might complicate things for non-winforms UI frontends though and might be overkill.

Assuming it even matters here, it could also put us down a bad path performance-wise. (I didn't dig super far but `RichTextBox` *does* seem to have a completely separate path for handling plain text.)

glopesdev:
@PathogenDavid good point, we've had issues with high-frequency streams before. We now buffer and batch updates automatically so the impact should be less pronounced but would be nice to have a performant code path by default.

I guess another option might be to just replace the control characters by their literal c# escape characters, i.e. tab with `\t`, newline with `\n`, etc.

@bruno-f-cruz any thoughts or feedback on this?

glopesdev:
@bruno-f-cruz  Yup, I am a bit torn on this one too. The first time I remember hitting this was with matrix representations, either from Math.NET or OpenTK where they would actually include new lines in the `ToString` method result. For that one it is definitely nicer to simply flatten the rows with a regular space separator (since each row is bracketed anyway).

[...]

PR #1765, opened 2024-04-30
Modify text visualizer to better handle newlines and unprintable control characters

PathogenDavid:
This change replaces `\n` (and other more esoteric Unicode newlines) with a space.

Carriage returns are dropped entirely (with the assumption they'll come with line feeds and that for the purposes of this visualization nobody will care about a bare `\r`.)

Other control characters (except `\t`) are replaced with � (the Unicode replacement character.)

Here's what that looks like with some test strings:

image

Strings are now processed in batches when possible to take advantage of reusing the temporary `StringBuilder`. This also allows us to skip processing of values which will never be displayed in the case of high-frequency inputs.

The design of this PR is such that future specializations of this visualizer (as mentioned in https://github.com/bonsai-rx/bonsai/issues/1670#issuecomment-2083904942) can simply override `MakeDisplayString` to change how the display string processing is done.

I also added disposing of `textPanel`, which was previously missing.

As an aside, I still implemented `Show(object)` since it's publicly accessible, but I did notice that the visualizer's actual text might not actually be updated in situations where someone was calling it directly and no batches come in to `ShowBuffer`. (This edge case wasn't handled before either.) Maybe the implementation should throw or assert to discourage using it? (Or if there's practical reasons to use it, maybe we should have it update the display anyway even though that'd be very inefficient.)

Fixes https://github.com/bonsai-rx/bonsai/issues/1670

### DIFF
--- a/Bonsai.Design/ObjectTextVisualizer.cs
+++ b/Bonsai.Design/ObjectTextVisualizer.cs
@@ -3,9 +3,10 @@ using System.Collections.Generic;
 using System.Windows.Forms;
 using Bonsai;
 using Bonsai.Design;
+using System.Diagnostics;
 using System.Drawing;
 using System.Reactive;
-using System.Text.RegularExpressions;
+using System.Text;
 
 [assembly: TypeVisualizer(typeof(ObjectTextVisualizer), Target = typeof(object))]
 
@@ -30,25 +31,99 @@ namespace Bonsai.Design
         /// <inheritdoc/>
         protected override void ShowBuffer(IList<Timestamped<object>> values)
         {
-            if (values.Count > 0)
+            if (values.Count == 0)
+                return;
+
+            var sb = new StringBuilder();
+
+            // Trim old values if bufferSize was reduced
+            while (buffer.Count >= bufferSize)
+                buffer.Dequeue();
+
+            // Add new values to the buffer (and only the ones which might appear)
+            for (int i = Math.Max(0, values.Count - bufferSize); i < values.Count; i++)
             {
-                base.ShowBuffer(values);
-                textBox.Text = string.Join(Environment.NewLine, buffer);
-                textPanel.Invalidate();
+                sb.Clear();
+                AppendDisplayText(sb, values[i].Value);
+                AppendString(sb.ToString());
             }
+
+            // Update the visual representation of the buffer
+            RefreshVisualization(sb);
         }
 
         /// <inheritdoc/>
         public override void Show(object value)
         {
-            value ??= string.Empty;
-            var text = value.ToString();
-            text = Regex.Replace(text, @"\r|\n", string.Empty);
-            buffer.Enqueue(text);
-            while (buffer.Count > bufferSize)
+            // Updates to this visualizer are expected to go through ShowBuffer
+            Debug.Fail($"Likely unintentional call to {nameof(ObjectTextVisualizer)}.{nameof(Show)}");
+
+            var stringBuilder = new StringBuilder();
+            AppendDisplayText(stringBuilder, value);
+            AppendString(stringBuilder.ToString());
+            RefreshVisualization(stringBuilder);
+        }
+
+        /// <summary>
+        /// Appends the display text for the specified object to the text buffer.
+        /// </summary>
+        /// <param name="stringBuilder">The string builder which receives the display text.</param>
+        /// <param name="value">The object for which to retrieve the display text.</param>
+        protected virtual void AppendDisplayText(StringBuilder stringBuilder, object value)
+        {
+            string rawText = value?.ToString() ?? string.Empty;
+            stringBuilder.EnsureCapacity(stringBuilder.Length + rawText.Length);
+
+            foreach (var c in rawText)
             {
-                buffer.Dequeue();
+                switch (c)
+                {
+                    // Carriage returns are presumed to be followed by line feeds, skip them entirely
+                    case '\r':
+                        continue;
+                    // Newlines become space so that things like multi-line JSON or matrices are still visible
+                    case '\n':
+                    case '\x0085': // Next line character (NEL)
+                    case '\x2028': // Unicode line separator
+                    case '\x2029': // Unicode paragraph separator
+                        stringBuilder.Append(' ');
+                        break;
+                    case '\t':
+                        stringBuilder.Append(c);
+                        break;
+                    // Replace all other control characters with the "�" replacement character
+                    case < ' ': // C0 control characters
+                    case '\x007F': // Delete
+                    case >= '\x0080' and <= '\x009F': // C1 control characters
+                        stringBuilder.Append('\xFFFD');
+                        break;
+                    default:
+                        stringBuilder.Append(c);
+                        break;
+                }
             }
+        }
+
+        private void AppendString(string value)
+        {
+            if (buffer.Count >= bufferSize)
+                buffer.Dequeue();
+
+            buffer.Enqueue(value);
+        }
+
+        private void RefreshVisualization(StringBuilder stringBuilder)
+        {
+            Debug.Assert(buffer.Count <= bufferSize);
+            stringBuilder.Clear();
+            foreach (var line in buffer)
+            {
+                if (stringBuilder.Length > 0)
+                    stringBuilder.Append(Environment.NewLine);
+                stringBuilder.Append(line);
+            }
+            textBox.Text = stringBuilder.ToString();
+            textPanel.Invalidate();
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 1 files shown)


Bonsai.Design/ObjectTextVisualizer.cs
```diff
@@ -30,27 +31,90 @@ public class ObjectTextVisualizer : BufferedVisualizer
...
+            var rawText = value?.ToString() ?? string.Empty;
+            var stringBuilder = new StringBuilder(rawText.Length);
+            MakeDisplayString(rawText, stringBuilder);
+            DisplayString(stringBuilder.ToString());
+        }
+
+        /// <summary>Converts the specified string for display in this visualizer</summary>
+        /// <param name="rawText">The raw text to be converted</param>
+        /// <param name="stringBuilder">The string builder which receives the display string</param>
+        protected virtual void MakeDisplayString(string rawText, StringBuilder stringBuilder)
```
glopesdev, line 78:
```suggestion
        /// <summary>
        /// Appends the display text for the specified object to the text buffer.
        /// </summary>
        /// <param name="stringBuilder">The string builder which receives the display text.</param>
        /// <param name="value">The object for which to retrieve the display text.</param>
        protected virtual void AppendDisplayText(StringBuilder stringBuilder, object value)
```

Since we are making this part of the API, I was considering we could refactor the method name to the above for consistency with `StringBuilder` methods and also for similarity with [`GetDisplayText`](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.design.collectioneditor.getdisplaytext) which we use elsewhere for similar purposes.

The idea behind passing here the actual `object` is to give more flexibility for future specialized text visualizers which might want to access specific instance fields or properties of the object directly, e.g. reflection, etc for debugging purposes. Looking at the code I was hoping it shouldn't impact performance too much since `StringBuilder` should resize efficiently in most cases.

Bonsai.Design/ObjectTextVisualizer.cs
```diff
@@ -30,27 +31,90 @@ public class ObjectTextVisualizer : BufferedVisualizer
...
+                        stringBuilder.Append('\xFFFD');
+                        break;
+                    default:
+                        stringBuilder.Append(c);
+                        break;
+                }
             }
         }
 
+        private void DisplayString(string value)
```
glopesdev, line 110:
```suggestion
        private void AppendString(string value)
```

This is just a nitpick since it is a private method, but when reading the code I was being thrown off by the name since the method doesn't actually display the string.

Bonsai.Design/ObjectTextVisualizer.cs
```diff
@@ -30,27 +31,90 @@ public class ObjectTextVisualizer : BufferedVisualizer
...
         {
-            value ??= string.Empty;
-            var text = value.ToString();
-            text = Regex.Replace(text, @"\r|\n", string.Empty);
-            buffer.Enqueue(text);
-            while (buffer.Count > bufferSize)
+            var rawText = value?.ToString() ?? string.Empty;
+            var stringBuilder = new StringBuilder(rawText.Length);
+            MakeDisplayString(rawText, stringBuilder);
+            DisplayString(stringBuilder.ToString());
```
glopesdev, line 72: Do we need to update the text box and invalidate the text panel here?

Although I agree we probably should just throw here. This code should never be called anyway now that `ShowBuffer` takes care of everything. The idea for buffered visualizers is by default they just call `Show` successively for each value, but since we have a more optimized implementation here this is not used.

Derived classes can just override it again if they need to do something with it.

Bonsai.Design/ObjectTextVisualizer.cs
```diff
@@ -30,27 +31,90 @@ public class ObjectTextVisualizer : BufferedVisualizer
...
+            for (int i = Math.Max(0, values.Count - bufferSize); i < values.Count; i++)
             {
-                base.ShowBuffer(values);
-                textBox.Text = string.Join(Environment.NewLine, buffer);
-                textPanel.Invalidate();
+                var rawText = values[i].Value?.ToString() ?? string.Empty;
+                sb.Clear();
+                sb.EnsureCapacity(rawText.Length);
+                MakeDisplayString(rawText, sb);
+                DisplayString(sb.ToString());
```
glopesdev, line 50:
```suggestion
                sb.Clear();
                AppendDisplayText(sb, values[i].Value);
                AppendString(sb.ToString());
```

Given all the above, I guess this bit would look like this. I was wondering here whether we will ever want to provide the timestamp, but I guess if we ever need that we can later make a non-breaking change by introducing a new virtual method `AppendTimestampedDisplayText` which receives the timestamped value and by default just calls `AppendDisplayText` with the value.

glopesdev (changes requested):
Looks great, just added a few refactoring suggestions for future-proofing more complex text visualizers and naming consistency.

Bonsai.Design/ObjectTextVisualizer.cs
PathogenDavid: I wasn't super happy with the name so this is much better, thanks!

I actually just moved the `EnsureCapacity` call to inside `AppendDisplayText` since that's really where it belonged anyway, so it really only leads to an extra allocation in the `Show(object)` case which we don't really care about.

Bonsai.Design/ObjectTextVisualizer.cs
PathogenDavid: Yeah, providing the timestamp to `AppendDisplayText` just opens another weird thing to handle in the `Show(object)` edge case.

I think if we're making a text visualizer that also shows timestamps then at that point we need to revisit making this visualizer have configuration.

Bonsai.Design/ObjectTextVisualizer.cs
```diff
@@ -30,25 +31,99 @@ public class ObjectTextVisualizer : BufferedVisualizer
...
+                buffer.Dequeue();
+
+            // Add new values to the buffer (and only the ones which might appear)
+            for (int i = Math.Max(0, values.Count - bufferSize); i < values.Count; i++)
             {
-                base.ShowBuffer(values);
-                textBox.Text = string.Join(Environment.NewLine, buffer);
-                textPanel.Invalidate();
+                sb.Clear();
+                AppendDisplayText(sb, values[i]);
```
glopesdev, line 47:
```suggestion
                AppendDisplayText(sb, values[i].Value);
```

To keep the current default value we need to get the value here to avoid printing the timestamp.

Bonsai.Design/ObjectTextVisualizer.cs
```diff
@@ -30,25 +31,99 @@ public class ObjectTextVisualizer : BufferedVisualizer
...
+            stringBuilder.EnsureCapacity(stringBuilder.Length + rawText.Length);
+
+            foreach (var c in rawText)
             {
+                switch (c)
+                {
+                    // Carriage returns are presumed to be followed by line feeds, skip them entirely
+                    case '\r':
+                        continue;
+                    // Newlines become space so that things like multi-line JSON or matricies are still visible
```
glopesdev, line 84:
```suggestion
                    // Newlines become space so that things like multi-line JSON or matrices are still visible
```

Bonsai.Design/ObjectTextVisualizer.cs
```diff
@@ -30,25 +31,99 @@ public class ObjectTextVisualizer : BufferedVisualizer
...
+                    case '\n':
+                    case '\x0085': // Next line character (NEL)
+                    case '\x2028': // Unicode line separator
+                    case '\x2029': // Unicode paragraph separator
+                        stringBuilder.Append(' ');
+                        break;
+                    case '\t':
+                        stringBuilder.Append(c);
+                        break;
```
glopesdev, line 93: Can we move this case next to `default`, or will it break the switch logic? Although I guess we could probably also leave it as it is for readability.

glopesdev (changes requested):
Just a few final corrections.

Bonsai.Design/ObjectTextVisualizer.cs
PathogenDavid: It needs to come before the control characters block since tabs are control characters. I think it's better to show the edge case is explicitly handled.

Bonsai.Design/ObjectTextVisualizer.cs
PathogenDavid: Oops, that's what I get for changing my test workflow. (I added a timer and thought the timestamps were what it was emitting, didn't notice they were on the other lines too.)

PathogenDavid:
Final corrections done. Let me know if you'd prefer I squash the two PR address commits together. (I normally would since them being separate isn't helpful, but I know it drives some people up the wall to have force pushes in a reviewed PR.)

glopesdev:
@PathogenDavid squashing and force pushing is fine 👍

PathogenDavid:
Done! (Ended up squishing it all down into a single commit since both addressing commits were pretty inconsequential.)

glopesdev (approved):

Issue #1645, opened 2023-12-12
BindTexture does not allow indexing dynamic TextureSequence


glopesdev:
The `BindTexture` operator currently allows indexing of texture sequences loaded from resources, but not dynamic texture sequences created with the `StoreImageSequence` operator.

PR #1770, opened 2024-05-03
Add support for binding textures directly from a `TextureSequence`

PathogenDavid:
Fixes https://github.com/bonsai-rx/bonsai/issues/1645

### DIFF
--- a/Bonsai.Shaders/BindTexture.cs
+++ b/Bonsai.Shaders/BindTexture.cs
@@ -136,7 +136,12 @@ namespace Bonsai.Shaders
             return Process(source, input =>
             {
                 GL.ActiveTexture(TextureSlot);
-                GL.BindTexture(TextureTarget, input != null ? input.Id : 0);
+                int textureId;
+                if (Index is int index)
+                    textureId = ((TextureSequence)input).Textures[index];
+                else
+                    textureId = input.Id;
+                GL.BindTexture(TextureTarget, textureId);
             });
         }
     }


Bonsai.Shaders/BindTexture.cs
```diff
@@ -136,7 +136,12 @@ public IObservable<Texture> Process(IObservable<Texture> source)
             return Process(source, input =>
             {
                 GL.ActiveTexture(TextureSlot);
-                GL.BindTexture(TextureTarget, input != null ? input.Id : 0);
+                int textureId;
+                if (Index is int index)
+                    textureId = ((TextureSequence)input).Textures[index];
+                else
+                    textureId = input.Id;
```
glopesdev, line 143:
```suggestion
                var textureId = Index is int index
                    ? ((TextureSequence)input).Textures[index]
                    : input.Id;
```

This is great, for some reason it never hit that you could use nullables against value pattern matching, but of course it makes complete sense.

For consistency with the code above I would maybe just change this to ternary operator form. In fact you could probably also change the code in line 82 to use value pattern matching as it should be more efficient.

PathogenDavid: Yup! Pattern matching is the best.

And done!

PathogenDavid:
Something I realized while applying pattern matching to the other path is that these two paths handle getting the texture ID at different times. (My path reads it in the `shader.Update` function, the other one reads it in the `Observable.Defer` and captures it for the closure passed to `shader.Update`.)

Not entirely sure if that has any observable behavior differences. My gut says that my implementation is probably slightly better since consumers of `Texture` probably shouldn't make assumptions about the ID since they don't own it, but let me know if you want me to make them match either direction.

glopesdev:
> My path reads it in the shader.Update function, the other one reads it in the Observable.Defer and captures it for the closure passed to shader.Update

This can be an important point, the main difference is that the timing of the code outside the closure can be controlled by the input sequence, whereas running code inside the closure we cannot make any assumptions about when exactly it will be run. This can be important when `Index` is changed dynamically just before the call is made, e.g. using `InputMapping`. Sometimes we do this to emulate batch execution of render pipelines.

Because of this detail, if anything I would tend to err on the side of capturing outside the closure. The easiest way to do this would be to change the signature of the delegate to receive the captured nullable, e.g.:

```c#
        IObservable<TSource> Process<TSource>(IObservable<TSource> source, Action<int?, TSource> update)
```

PathogenDavid:
> Because of this detail, if anything I would tend to err on the side of capturing outside the closure. The easiest way to do this would be to change the signature of the delegate to receive the captured nullable, e.g.:

That's sensible. Looking at it closer though, I ended up refactoring the function to take a getter for the texture ID instead of an update function.

Your suggestion would've only unified the capture of `Index`, but the actual texture ID was still different.

It's a bit more of a change, but I think it better matches the existing behavior exactly in both cases.

Bonsai.Shaders/BindTexture.cs
```diff
@@ -58,35 +58,25 @@ public bool IndexSpecified
             get { return Index.HasValue; }
         }
 
-        IObservable<TSource> Process<TSource>(IObservable<TSource> source, Action<TSource> update)
+        /// <param name="source">The source sequence to be processed</param>
```
PathogenDavid, line 61: This is really only here to make [CS1573](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs1573) happy. (Unfortunately it cares even if the method is private like this one.)

I'm sort-of not a fan of the XML documentation warnings because they encourage you to create useless self-evident comments like this.

Bonsai.Shaders/BindTexture.cs
```diff
@@ -133,10 +144,11 @@ public IObservable<TSource> Process<TSource>(IObservable<TSource> source)
...
         public IObservable<Texture> Process(IObservable<Texture> source)
         {
-            return Process(source, input =>
+            return Process(source, (maybeIndex, _, input) =>
             {
-                GL.ActiveTexture(TextureSlot);
-                GL.BindTexture(TextureTarget, input != null ? input.Id : 0);
+                return maybeIndex is int index
+                    ? ((TextureSequence)input).Textures[index]
+                    : input.Id;
```
glopesdev, line 151: If I am reading this correctly, I think there is another important sense in which this refactor breaks existing behavior. The intended semantics for the operator is the following:

1. If a texture name is specified, that texture is used, regardless of the type of the `source` sequence.
2. If no texture name is specified, but the type of the input is `Texture`, then the elements of `source` are used for binding.

The reason for giving primacy to texture name is to avoid silent changes in behavior when the type of the input source is "accidentally" a `Texture`. As a concrete example, imagine a situation where we need to bind both a dynamic texture and a static texture for a shader with two texture uniforms:

image

In the first `BindTexture` we would leave `TextureName` empty so we can bind the dynamic texture to `Texture0`, but in the second `BindTexture` we would specify the name of a static texture we want to bind to `Texture1`. In other words, in the first node we care about the value of the input, whereas in the second node we care only about the timing of the event, i.e. we want to bind the second static texture right after we bind the dynamic texture.

In the previous implementation we could have it both ways, which is something we would like to keep at least for a patch release, although I do agree with you there was an inconsistency in the way that both inputs were captured in the closure that would be nice to resolve.

glopesdev (changes requested):

Bonsai.Shaders/BindTexture.cs
glopesdev: That's odd, on my vanilla VS community this doesn't get flagged as a warning somehow. If there are some editor settings we can toggle to turn this off it would be best, we really don't want to enforce having XML documentation comments for private method parameters.

PathogenDavid: I tested on Visual Studio 2022 in a fresh root suffix (so completely default settings) and pinned the .NET SDK to either 6.0 or 8.0 stable using global.json and it still happens.

I want to emphasize it's this specific line and that it's CS1573 rather than [CS1591](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1591).

I *did* want to document `getTextureId` despite it being private because bare `Func`/`Action` delegates tend to get kinda hairy when they have many parameters since there's no parameter names. Plus private XML documentation can still be useful since it shows up in Intellisense:

image

Bonsai.Shaders/BindTexture.cs
```diff
@@ -110,7 +100,28 @@ IObservable<TSource> Process<TSource>(IObservable<TSource> source, Action<TSourc
         /// </returns>
         public IObservable<TSource> Process<TSource>(IObservable<TSource> source)
         {
-            return Process(source, update: null);
+            var texture = default(Texture);
+            var textureName = default(string);
```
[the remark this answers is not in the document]
PathogenDavid: Yeah, I think I had it in my head these were declared outside the `Observable.Defer`, will revisit.

Bonsai.Shaders/BindTexture.cs
PathogenDavid: You're right, not sure how I overlooked that. Will fix!

PathogenDavid:
Revisiting this again with a fresh mind, I realized we were both getting a little caught up on the `update` delegate approach of the previous implementation.

The latest iteration does away with the delegate entirely and instead has `Process<TSourcce>` handle the input being a texture when the texture name isn't specified.

This technically makes the `Process(IObservable<Texture>)` overload redundant, but I kept it so that it's still documented separately since I felt like that probably makes more sense from a user perspective. (Can remove and revise the documentation if you feel it's unnecessary though.)

The diff looks like I changed documentation but I actually just moved the body of the private `Process(IObservable, Action)` into the body of `Process<TSource>` since the helper is no longer useful after removing the `Action`.

As an aside, is it desirable that we just nothing when there's no texture? Seems like it should maybe be an error.

Bonsai.Shaders/BindTexture.cs
```diff
@@ -76,43 +92,25 @@ IObservable<TSource> Process<TSource>(IObservable<TSource> source, Action<TSourc
                                 : null;
                         }
 
+                        texture ??= input as Texture;
```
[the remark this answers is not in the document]
PathogenDavid: > I see one potentially problematic edge-case: if you didn't specify a texture name, and you pass a sequence of valid textures, followed by a null texture, the `texture` state variable will bind to the latest non-null texture.

`BindTexture`: The most nuanced operator in all of Bonsai.

> (will this be optimized by the compiler when generic type will always fail the test? or the other way around, are we paying the cost of boxing to object by using as with generic types in cases where the actual type is a value type?)

For reference types it's implementation-defined whether the runtime will specialize a generic method for each `T` or if it will perform generic sharing. Not sure if it would in this case, my gut says it wouldn't since historically the JIT has been pretty conservative about doing so. (I'd check but Disasmo only works with modern .NET and it seems overkill to go mucking with WinDbg for this.) Generic sharing should always be avoided when inlining happens, but inlining almost certainly won't happen in this case since the method is fat and tail calls make more sense.

For value types it will always specialize each T out of necessity. (The .NET VM makes strong guarantees about how value types are treated that inevitably require it.)

It's mostly a moot point though, `input` will only be inspected when `texture` is `null`, it's only `null` if the node isn't configured. I don't think it's valuable to sweat the small stuff for a `BindTexture` node which isn't configured and doesn't take `Texture` elements. (The only reasonable case I can think of this happening is if `TextureName` is externalized and someone sets it to `null`, but this doesn't actually do anything useful since the unbinding is skipped in that situation.)

> on the other hand, if we know the type is a texture we could just do a direct type-cast rather than `as`.

This doesn't ever have any meaningful performance difference as far as I'm aware, they're very similar operations because both have to handle the cast being wrong. (And that check would only be elided if the generic was specialized, which would've also removed the similar handling for `as` anyway.)

> getter delegate might be faster

The getter delegate has the overhead of allocating the closure on the heap. Invoking it isn't gonna be super friendly towards the data/instruction cache or branch predictor. I'd be shocked if it was consistently faster over casting :)

-----------------

Anyway your code as suggested wasn't going to fix the missing handling of `null`-to-unbind.

I pushed a new iteration (hopefully the final one lol) which properly handles `null` in the texture-from-input case and avoids incorrectly caching the input texture.

Bonsai.Shaders/BindTexture.cs
```diff
@@ -58,35 +58,40 @@ public bool IndexSpecified
...
+                            cachedTextureName = TextureName;
+                            cachedTexture = !string.IsNullOrEmpty(cachedTextureName)
+                                ? shader.Window.ResourceManager.Load<Texture>(cachedTextureName)
                                 : null;
                         }
 
-                        if (texture != null)
+                        var texture = cachedTexture ?? input as Texture;
+
+                        if (texture != null || allowUnbind)
```
glopesdev, line 81: We actually don't want the unbind behavior for this patch, as that would be a breaking change. I agree the nitpick about the performance of the `as` operator is most likely irrelevant, so probably the following would be enough in a single method:

```c#
var activeTexture = texture ?? input as Texture;
if (activeTexture != null)
{
    var textureId = Index is int index
        ? ((TextureSequence)activeTexture).Textures[index]
        : activeTexture.Id;
    shader.Update(() =>
    {
        GL.ActiveTexture(TextureSlot);
        GL.BindTexture(TextureTarget, textureId);
    });
}
```

I opted to create a new local rather than rename the cache variable to minimize diffs, but that is just a style preference.

glopesdev (changes requested):
Sorry for the confusion in my previous comment, it took me a while to fully reload the context of this operator and while doing several edits to the comment I probably failed to explain myself correctly.

I think this will be the last iteration and happy to go back to having everything in one method. This is basically the same as your previous proposal except introducing an extra local instead of renaming the cache variable.

Indeed this operator is surprisingly nuanced, and I learned several things from this review so definitely worthwhile!

Bonsai.Shaders/BindTexture.cs
PathogenDavid: The current implementation supports this unbind behavior:

https://github.com/bonsai-rx/bonsai/blob/ced27db65e266c0eab41a41f55cb22b70ce553b6/Bonsai.Shaders/BindTexture.cs#L139

Also IMO it's better for the "special" variable to have the unique name. It'd be easy to overlook someone accidentally using `texture` instead of `activeTexture` down below. It also makes it more clear what that closed variable is actually for and avoids mistakes like the one I made earlier with caching the `input`.

I definitely get wanting to avoid diff churn, but things should be optimized for people reading the current code not its history.

glopesdev: > The current implementation supports this unbind behavior

Ah, you are completely correct! In that case let's leave it then.

Completely agree on optimizing for reading, my main reason for trying to minimize diffs is making it easier to use git blame tools. Anyway, in this case not critically important.

glopesdev (approved):

Issue #1507, opened 2023-08-02
Exception thrown when cross-reference file link returns invalid format


glopesdev:
Currently the editor tries to fetch a `xrefmap.yml` file with all documentation references if the project package specifies a top-level URL. If this URL points to something other than a valid YML file the editor will throw an exception during deserialization.

We will want to ensure that an actual YML file has been downloaded (and not for example an HTML redirect page), and consider a different way of handling deserialization errors other than the default unhandled exception dialog.

PR #1785, opened 2024-05-09
Fixed uncaught exception when XRefMap is invalid

PathogenDavid:
The deserialization exceptions were making it to the generic unhandled exception dialog due to the relevant catch clause looking for `SystemException`.

`SystemException` (and its complement `ApplicationException`) is a bit of a failed experiment from .NET Framework 1. It's no longer used for new `System.*` exceptions, and even before .NET Core it was applied very inconsistently. IMO the line between exceptions in the `System` namespace and exceptions outside of it is basically meaningless and very fuzzy regardless.

As such this PR increases the scope to just plain `Exception`.

------------------

This still would have resulted in very vague low-level errors reaching the end user, so I introduced `DocumentationException` to denote specific errors when parsing the XRefMap.

When one XRefMap fails due to a 404 and another fails due to a `DocumentationException`, the `DocumentationException` takes priority.

------------------

Here is the error that occurs when the documentation server responds with something other than an XRefMap:

image

And here's the error that occurs when the documentation server responds with something that claims to be an XRefMap but isn't unusable:

image

Fixes https://github.com/bonsai-rx/bonsai/issues/1507

### DIFF
--- a/Bonsai.Editor/DocumentationHelper.cs
+++ b/Bonsai.Editor/DocumentationHelper.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Net;
 using System.Net.Cache;
 using System.Threading.Tasks;
+using YamlDotNet.Core;
 using YamlDotNet.Serialization;
 using YamlDotNet.Serialization.NamingConventions;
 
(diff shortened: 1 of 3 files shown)


Bonsai.Editor/DocumentationHelper.cs
```diff
@@ -61,11 +61,23 @@ static async Task<Uri> GetDocumentationAsync(string baseUrl, string uid)
...
             var deserializer = new DeserializerBuilder()
                 .WithNamingConvention(CamelCaseNamingConvention.Instance)
                 .IgnoreUnmatchedProperties()
                 .Build();
-            var xrefmap = deserializer.Deserialize<XRefMap>(reader);
+            XRefMap xrefmap;
+            try
+            { xrefmap = deserializer.Deserialize<XRefMap>(reader); }
+            catch (YamlException ex)
+            { throw new DocumentationException("The cross-reference map returned by the documentation server is malformed.", ex); }
```
glopesdev, line 76:
```suggestion
            try { xrefmap = deserializer.Deserialize<XRefMap>(reader); }
            catch (YamlException ex)
            {
                throw new DocumentationException("The cross-reference map returned by the documentation server is malformed.", ex);
            }
```

For same style reasons as above.

glopesdev (changes requested):
This looks great, thanks! Just very minor style tweaks for consistency with current code, looking forward to drafting together a style guide for modern .NET.

PathogenDavid:
Fixed!

> looking forward to drafting together a style guide for modern .NET.

I'll definitely have strong opinions about your `try { SingleLine(); }` habit. It breaks auto-formatting 😅

I can cope for now though.

glopesdev (approved):

Issue #1769, opened 2024-05-02
Stack overflow exception from VisualizerMapping with second VisualizerMapping as input


RoboDoig:
While investigating #1591 I noticed that creating this workflow is possible in the editor:

![image](https://github.com/bonsai-rx/bonsai/assets/11314305/9a57567f-c61c-4249-be9d-e5cbc8a46936)

When run this results in a terminating crash from a stack overflow exception, presumably due to a cyclical dependency.

RoboDoig:
Looked into this a bit more, the issue comes from entering an infinite recursive call in `LayoutHelper.GetMashupArguments`. For example in this case, the `Int` mapping source argument will be passed on every recursive call.

I can't think of any situation where you would want to combine visualizers as in the workflow example above as the information is redundant. In that case the fix may be just to disallow visualizer combinations where the data source is the same for visualizers in the combination.

PR #1787, opened 2024-05-10
Prevent infinite recursion when visualizer mappings of the same node are combined together

PathogenDavid:
This is potentially not the most elegant solution possible here, but Gonçalo asked me to take a quick look and see if we could squeeze a fix in for 2.8.3.

Maybe this is good enough to just prevent the crash since we don't expect anyone to actually intentionally make graphs like this and this code only runs on startup. If not we could leave https://github.com/bonsai-rx/bonsai/issues/1769 open and do a more thorough investigation later.

I'm not attached to the wording of the message if anyone has a better suggestion. (Might be more accurate to say the same node appears multiple times in the mashup? Not sure if people would actually consider `VisualizerMapping` associations to be mashups though, that feels like internal nomenclature leaking to the surface.)

For some reason the error dialog shows up twice, but I think that's actually an unrelated bug. (The exception is only actually thrown once, at first I thought it was getting thrown for each `VisualizerMapping` node but I verified that's not the case.)

Fixes https://github.com/bonsai-rx/bonsai/issues/1769

### DIFF
--- a/Bonsai.Editor/Layout/LayoutHelper.cs
+++ b/Bonsai.Editor/Layout/LayoutHelper.cs
@@ -193,6 +193,10 @@ namespace Bonsai.Design
             if (visualizerMappings.Count == 0) return Array.Empty<VisualizerFactory>();
             return visualizerMappings.Select(mapping =>
             {
+                // stack overflow happens if a visualizer ends up being mapped to itself
+                if (mapping.Source == builder)
+                    throw new WorkflowBuildException("Combining together visualizer mappings from the same node is not currently supported.", builder);
+
                 var nestedSources = GetMashupArguments(mapping.Source, typeVisualizerMap);
                 var visualizerType = mapping.VisualizerType ?? typeVisualizerMap.GetTypeVisualizers(mapping.Source).FirstOrDefault();
                 return new VisualizerFactory(mapping.Source, visualizerType, nestedSources);


glopesdev:
@PathogenDavid I was looking over this again and I think the infinite recursion only happens in the specific case where the same mapping is being called over itself, in which case we can detect it by just testing for `mapping.Source == builder`.

I believe this to be the case since we know the graph is a DAG by definition, so no real cycles. I believe the only reason we get a call cycle here must therefore be by propagation of the same builder onto itself, which can only happen under the small set of no-modification rules, and crucially, it happens only for the same builder reference. As soon as that builder is modified by some interaction with other nodes, it becomes a different builder, and therefore overlaying that on top of the original builder is completely fine.

If this turns out to be true, then it also raises the question of whether we should throw at all, or simply ignore or better handle this specific case.

The motivation to ignore would be that overlaying the exact same visualizer onto itself (with exactly the same data) would always be the same exact computation. We could consider such operation to be idempotent and therefore simply ignore the recursion, e.g. in `GetMashupArguments` do:

```c#
            return visualizerMappings.Where(mapping => mapping.Source != builder).Select(mapping =>
```

The only slightly strange interpretation here the case where we map the exact same builder onto itself but using a different visualizer type (i.e. map the same image onto itself as text for visualization purposes). Under this proposal, this stacking would simply be ignored, which might be unexpectedly weird. However, I feel in this case the right answer would not be to throw either, but rather generate some new visualizer which is really the compound of both.

PathogenDavid:
> in which case we can detect it by just testing for `mapping.Source == builder`.

Sounds good to me, the `HashSet` recursion detection was out of an abundance of caution in case there was a more complex setup we weren't seeing.

> If this turns out to be true, then it also raises the question of whether we should throw at all, or simply ignore or better handle this specific case.

We definitely can't just ignore it. @RoboDoig found that with the following workflow the data mashup basically just doesn't happen (the point is gone) if you just ignore the "cycle": [PointOnImage-IssueWithVisualizerMappingCycle.zip](https://github.com/bonsai-rx/bonsai/files/15307356/PointOnImage-IssueWithVisualizerMappingCycle.zip)

PointOnImage-IssueWithVisualizerMappingCycle

I think we talked about this during our meeting and IIRC you thought this was partially/wholly caused by https://github.com/bonsai-rx/bonsai/issues/1789, but testing it with a timer (as shown above) has the point being lost entirely if the bottom visualizer mapping is enabled. (The camera view actually updates live and the https://github.com/bonsai-rx/bonsai/issues/1789 problem doesn't even happen in the first place.

The current tip of this PR's branch has a preprocessor toggle to toggle between skipping the recursion and throwing if you want to poke at it. (I personally still need to dig in further to get a better view of the big picture here.)

---------------

Another reason we might want to throw for now is so we can more easily change the semantics of this currently-odd case in the future. For example, to handle the intentional self-mashup scenario we were talking about during dev club yesterday. (IE: The mapping a camera capture's info onto its image and such.)

glopesdev:
> Another reason we might want to throw for now is so we can more easily change the semantics of this currently-odd case in the future.

This is a good point, maybe in that case let's just throw with the simple test for now and we can expand the semantics for the next feature release.

PathogenDavid:
Cleaned up the branch to be the throwing implementation. Was slightly unsure on the exception message since as far as I know we never use the "mashup" wording in documentation and such, aimed for something that hopefully makes sense to anyone unfortunate enough to encounter this without thinking anything is wrong.

Bonsai.Editor/Layout/LayoutHelper.cs
```diff
@@ -193,8 +193,13 @@ static IReadOnlyList<VisualizerFactory> GetMashupArguments(InspectBuilder builde
             if (visualizerMappings.Count == 0) return Array.Empty<VisualizerFactory>();
             return visualizerMappings.Select(mapping =>
             {
-                var nestedSources = GetMashupArguments(mapping.Source, typeVisualizerMap);
+                // mapping.Source == builder if two visualizers in the mashup visualize the same node
+                var nestedSources = mapping.Source == builder
+                    ? throw new WorkflowBuildException("Visualizer mappings of same node are combined together.", builder)
+                    : GetMashupArguments(mapping.Source, typeVisualizerMap);
+
```
glopesdev, line 200:
```suggestion
                // stack overflow happens if a visualizer ends up being mapped to itself
                if (mapping.Source == builder)
                    throw new WorkflowBuildException("Combining together visualizer mappings from the same node is not currently supported.", builder);
                
                var nestedSources = GetMashupArguments(mapping.Source, typeVisualizerMap);
```

For future refactoring and emphasis I would separate the detection into its own section. You are right about getting the right wording, I am not entirely sure myself how to make this less confusing, and ultimately there might not even be a good way to do this.

I ended up converging on something that ends with "not currently supported." as a way of hinting that this is something weird that we don't know whether it makes sense or not. The reason it might be helpful to hint this is so people reach out to us with a potentially really useful case that might motivate us to resolve this better.

glopesdev (approved):

PR #1795, opened 2024-05-11
Fix race condition within `HandleWorkflowError`

PathogenDavid:
This was noticed while Gonçalo and I were investigating https://github.com/bonsai-rx/bonsai/issues/1796

Also adds a more sensible caption to the error dialog when non-workflow errors occur.

I considered adding a message to the error encouraging people to report the error since we don't expect bare non-workflow exceptions to escape to this point, but the part of my brain responsible for wording things is already running on fumes as I get all these issues out of my head before the weekend 😅

### DIFF
--- a/Bonsai.Editor/EditorForm.cs
+++ b/Bonsai.Editor/EditorForm.cs
@@ -1407,6 +1407,7 @@ namespace Bonsai.Editor
 
         void HandleWorkflowError(Exception e)
         {
+            var building = this.building; // Ensure the current value of `building` is used when if the handler below is invoked
             Action selectExceptionNode = () =>
             {
                 var workflowException = e as WorkflowException;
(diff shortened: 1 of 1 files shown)


Bonsai.Editor/EditorForm.cs
```diff
@@ -1415,7 +1416,7 @@ void HandleWorkflowError(Exception e)
                     workflowError = workflowException;
                     HighlightExceptionBuilderNode(workflowException, building);
                 }
-                else editorSite.ShowError(e.Message, Name);
+                else editorSite.ShowError(e.Message, "Internal workflow error");
```
glopesdev, line 1419:
```suggestion
                else editorSite.ShowError(e.Message, "Unexpected Exception");
```

I would maybe just keep the caption general, since I'm not sure we can guarantee at this point whether these will be only internal errors.

PathogenDavid:
Converting to a draft for now as this shouldn't be merged. Will leave a more detailed comment on https://github.com/bonsai-rx/bonsai/issues/1796.

PathogenDavid:
Closing out stale PR. As noted in https://github.com/bonsai-rx/bonsai/issues/1796#issuecomment-2109886901 this wasn't the right approach, and the underlying issue was fixed by https://github.com/bonsai-rx/bonsai/pull/2212.

PR #1848, opened 2024-06-13
Show visualizer if checked menu item is selected but visualizer is hidden

### LINKED ISSUE #1841: Consider setting the Show Visualizer context menu selection to None on visualizer window close
When a user closes a visualizer window by clicking the X in the top right corner, the Show Visualizer option in the context menu doesn't change. My initial thought was that the context menu should update to Show Visualizer > None, but then I found that the visualizer settings persist because the Visible flag in the layout file gets set to false, while setting the Visualizer explicitly to None erases these settings entirely. 

Related to #1823 and perhaps does not need its own issue, but at least for me I find it confusing to exit out of the visualizer window and still see that the visualizer is selected. Double clicking on the node or changing selection can make it visible again, but maybe a simple fix would just be to always set Visible -> true on selection in context menu.

ncguilbeault:
This PR modifies the visualizer menu item event handler to display the visualizer if the menu item is checked and if the visualizer is hidden.

Fixes #1841

### DIFF
--- a/Bonsai.Editor/GraphView/WorkflowGraphView.cs
+++ b/Bonsai.Editor/GraphView/WorkflowGraphView.cs
@@ -1719,6 +1719,15 @@ namespace Bonsai.Editor.GraphView
                     }
                     UpdateSelection(forceUpdate: true);
                 }
+                else if (menuItem.Checked)
+                {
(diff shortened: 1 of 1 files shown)


glopesdev:
Superseded by #2158

PR #1870, opened 2024-07-02
Rewrite editor navigation model

### LINKED ISSUE #1823: Proposal for new editor navigation model
The current editor navigation model has proven unwieldy for large, deeply nested, workflows.

This proposal is for a simplification of the navigation model which should be more scalable, and hopefully require minimal modifications to the current implementation.

### Basic Principles

- Only one workflow graph view at a time is visible and active in the canvas;
- Navigating to a nested workflow switches the view to render the new workflow;
- Breadcrumbs or tree view control assists navigation across the hierarchy.

Essentially the analogy here is to a browser, where navigating hyperlinks to other pages simply opens and renders the new page in the same window by default.

There are many implications, opportunities for redesign, and performance optimisations, that can result from this proposal. Below we try to collect and describe them in more detail as we analyse the proposal.

### Render Logic

The `GraphNode` class is the backbone for representation of workflows in the editor controls. `GraphNode` objects are currently stored and manipulated in many places, including the undo/redo stack and visualizer associations. The complexity of keeping track of this joint state and the associations between graph nodes and expression tree nodes have been a frequent source of bugs and tangled logic.

This proposal is an opportunity to remove this representation strategy entirely, and adopt a purely functional approach where graph nodes are created and used simply for the current render layout state, and discarded as soon as the structure changes.

### Visualizers

Visualizers were previously associated with editor view graph nodes, but in this model it might be simpler to associate them directly with the expression builder nodes, using a global map.

This will make it easier to have all visualizers be children of the main window, which will resolve the issue of nested workflow editor windows having to be open to display deep visualizers.

### Error Highlighting

Both compiler and runtime errors were previously reported by opening nested editor windows in order to reveal the innermost cause of unhandled exceptions. This was clunky as it led to a potential cascade of new editor windows.

Under this proposal, we can reimagine error highlighting by leveraging the tree view, and simply place a mark next to the hierarchy levels which are reporting errors. Opening the corresponding level will highlight any error nodes contained within.

### Include Workflows

How should `IncludeWorkflow` operators be displayed in the tree view? Currently we were aiming to strike a balance between using them as building blocks (no need to navigate into them) and organization (components of a larger system which may be navigated into quite often).

### Disabled Workflows

How should disabled group workflows be handled? A few options:
- Handle disabled groups as a primitive non-group node, so you cannot explore them
- Render the internals as either all nodes disabled, or with some kind of annotation marking the whole workflow as disabled
- Limit expansion the same way as include workflows (require explicit Go To Definition) and make them read-only

### Undo/Redo Stack

Currently navigating into or out of a nested workflow was being logged as an undo/redo command. Under this proposal it might make more sense to have only changes to the workflow be logged in the stack, however it is still required to show the undo/redo changes in the context of the workflow level which is changing. This should be straightforward assuming there is always a way to immediately navigate to the workflow which is currently being modified.

### Navigation paths

Related to both the above and the serialization of visualizer layout files and visualizer mashups, there is a possibility of specifying "navigation paths" which would allow selecting or opening any specific node of the workflow at any nesting level. Such paths might be serialized in the `.layout` file, allowing selective serialization of only those visualizers which are actually open or assigned, rather than the current strategy where all visualizers of active layers are stored, even if they have never been opened.

It would have been nice here to include the option for "querying" the node by name. Unfortunately this is hard since we have currently no constraints whatsoever on node names, and it would make the parser too complex.

One option to work with non-sanitized paths existing right now would be to treat any group node names with invalid characters as if they had no name, e.g. `Task Controller (SelectMany)` would be considered to have the name `SelectMany`.

Only group node names should be included in navigation paths, i.e. node names coming from `INamedElement` are not allowed. The only exception might be `IncludeWorkflow` operators where we might want to use the name of the included operator, assuming it is sanitized.

Draft proposals for "ideal" navigation path:

#### Purely indexed paths
```
0.3.23
```

#### Hybrid name + index
```
#Hardware.SelectMany[3].Behavior[2]
```

#### Indices only in leaf nodes
```
#Hardware.TaskController.Behavior[2]
```

### LINKED ISSUE #1993: Consider expanding explorer level on clicking the node text
The tree can be expanded by clicking the '+'. This is ok but it tends to be a bit hard given the small size of the button. It would be interesting to prototype an interaction where the menu would expand after the group is selected instead of having to click on the plus sign

### LINKED ISSUE #2000: Breadcrumb buttons clipping off container at some DPI scales
This seems to be related to container size. To reproduce run the editor with no DPI scaling. It might be possible that required resizing might vary with the scale factor, so we should perhaps consider overriding `ScaleControl` or turning on `AutoSize`.

### LINKED ISSUE #2001: Switching visualizers at runtime opens an extra out of synch visualizer
To reproduce, insert any infinite source with two visualizers, open the default visualizer, and finally switch to the text visualizer. The default visualizer will close and reopen and it won't be automatically closed on stopping the workflow.

### LINKED ISSUE #2004: Accessing a disabled nested operator throws an error
To reproduce:
1. Add a GroupWorkflow
2. Disable
3. Double click to open

Depending on the complexity of the workflow it sometimes even leads to a deadlock of the editor (cant seem to reproduce this reliably tho)

glopesdev:
### Workflow Explorer
  - Add workflow explorer treeview
  - Allow navigating to any level without opening previous levels
  - Allow visualizers to run uncoupled from the workflow view
  - Highlight build errors in explorer treeview

### Dock Panel
- Add support for multi-panel navigation
- Allow tab navigation across panels and controls
- Add support for changing dock panel theme
- Add support for persisting dock panel layout

Fixes #1823 
Fixes #1993 
Fixes #2000 
Fixes #2001 
Fixes #2004


glopesdev:
Early feedback from the preview is that allowing opening of independent tabs and windows should be allowed in the editor control to allow watching different parts of the workflow.

We could take this opportunity to generalize tabs and floating windows into a single dock panel. The main concern is going to be how to handle multiple windows pointing to the same path, and what to do when deleting group nodes causes invalidation of multiple open tabs.

For the first issue we could consider enforcing that windows are singleton w.r.t. a unique path, which would obviate having to synchronize multiple view states of the same workflow. For the second issue we could probably store it in the undo/redo stack as we did in the previous navigation model by closing as a command and scheduling reopening of tab/window pointing to the deleted path with same settings in the undo command.

glopesdev:
Feedback from DCM:
  - Changes in nested node names need to propagate to dock contents containing that node
  - Allow open in new window and open in new tab in context menu for multi-selection
  - Check for tolerance to errors if there is a mismatch between persisted dock state:
    - On error, invalidate everything
    - On error, try to cancel opening that dock content:
      - need to check how to do this in dock panel suite, may require dummy empty content to prune afterwards
    - On error, set the dock content to the workflow root
  - We may want to save editor layout when editor closes, regardless of whether the workflow was saved
  - Should editor layout and visualizer layout be saved at the same time, e.g. when editor closes:
    - Probably yes
  - Check whether visualizer layout is tolerant to errors:
    - On error, try to invalidate only the assignments that failed

Issue #1856, opened 2024-06-24
Improve release automation pipeline for core repository and docs


glopesdev:
Right now releasing new versions of Bonsai is a multi-hour process with lots of manual steps. In Bonsai 2.8.4 we'd like to move towards automating this release process and improve the robustness of our continuous integration (CI) efforts in general.

At a high level, this means that the following will start being automated:

* Determining which packages have changed and will need to be published
* Building and publishing packages to NuGet.org
* Creating the installer and portable zip and publishing them to GitHub
* Publishing the new version number and download links to the Bonsai website
* Updating the docs website
* Other miscellaneous release chores like closing milestones and bumping version numbers

Additionally we hope to design this new infrastructure in a way that allows adding release automation to other Bonsai packages outside of the main repo. (EG: Bonsai.Arduino)

These changes also represent some other improvements to Bonsai in general:

* Bonsai is will be able to be built under the modern .NET SDK (first steps towards .NET 8)
* Continuous integration testing is now performed for Linux
* Source Link is enabled (this is a huge benefit to developers of Bonsai packages, also means Bonsai packages and binaries are tagged with the specific Git revision they were built from)
* Deterministic builds
* General build system cleanup and fixes

PathogenDavid:
@glopesdev Fleshed out the issue description.

-----------------------

As per our meeting after dev club, how do we want to handle releasing packages only temporarily changed for unstable builds and preview versions of Bonsai?

After writing all this up I think I'm much more firmly in the "do nothing, assume this won't happen" camp, but I said I'd leave a comment for you so here it is :)

Determining which packages to release is done by comparing the packages between the current revision and the revision of the most recent stable release of Bonsai.

This creates an odd edge case for unstable and preview releases though. Imagine you have the following sequence of events:

* A: Release Bonsai 1.0.0
* B: Modify Bonsai.Dsp
* C: Modify Bonsai.Shaders
* D: Release Bonsai 1.0.1-alpha0
* E: Modify Bonsai.Dsp
* F: Revert change to Bonsai.Shaders
* G: Modify Bonsai.Core
* H: Release Bonsai 1.0.1-alpha1
* I: Release Bonsai 1.0.1

(The "Release" items are releases created on GitHub. The others are separate commits.)

For this sequence of commits, the following packages are published for Bonsai.Shaders and Bonsai.Dsp :

| Event | Bonsai.Shaders | Bonsai.Dsp |
|-----|-----|-----|
[...]

Issue #1831, opened 2024-05-31
Bonsai 2.8.3.4546 still reports itself as 2.8.2 in Help>About Menu


banchan86:

PathogenDavid:
This is happening because the about box specifically shows the version of `Bonsai.Editor`, but `Bonsai.Editor` didn't have any changes for 2.8.3 so we didn't release a new version of it.

We should probably either show the bootstrapper version there instead or always release a new version of `Bonsai.Editor` regardless.

It might also be a good idea to have the bootstrapper require a specific version of `Bonsai.Editor`, I feel like the minimum version match between the two only has the potential to cause problems. (Might be a good idea for all of the fundamental core Bonsai packages in general.)

banchan86:
Ah I see! Yea it might probably be better to update `Bonsai.Editor` as well so users aren't confused as to whether they had the latest version installed (as I did) haha.

PathogenDavid:
This will be fixed in 2.8.4, thanks for the report!

Issue #1849, opened 2024-06-13
Consider improving experience with authenticated package feeds


PathogenDavid:
(Mostly filing this issue for posterity, I don't think this is an even remotely high priority unless we start recommending GitHub Packages as a way for people to publish Bonsai packages.)

While working on release automation, I wanted to quickly test if Bonsai even worked with authenticated package feeds since GitHub Packages feeds always require authentication.

It works just fine if you manually add the credentials to `packageSourceCredentials`, but the user experience leaves a bit to be desired.

Right now attempting to enumerate an authenticated feed results in a somewhat generic error message:

![Screenshot of Bonsai's package manager with an error](https://github.com/bonsai-rx/bonsai/assets/278957/2cf2ccdc-823c-40a1-83ec-d5eb231e671c)

> There was an error searching the repository 'Bonsai Unstable'.
> --> Unable to load the service index for source https://nuget.pkg.github.com/bonsai-rx/index.json
> --> Response status code does not indicate success: 401 (Unauthorized).

The user can fix this by adding their credentials to their `NuGet.config`. IE:

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <!-- ... -->
    <add key="Bonsai Unstable" value="https://nuget.pkg.github.com/bonsai-rx/index.json" />
  </packageSources>
  <packageSourceCredentials>
    <Bonsai_x0020_Unstable>
      <add key="Username" value="GITHUB_USERNAME" />
      <add key="ClearTextPassword" value="GITHUB_PERSONAL_ACCESS_TOKEN" />
    </Bonsai_x0020_Unstable>
  </packageSourceCredentials>
</configuration>
```

As hinted above, in the case of a GitHub Packages feed the user would need to [create a personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#creating-a-personal-access-token-classic) to use as the password.

----------------------------------------

Possible solutions:

1. Detect a 401 and link to this issue
2. Same as 1, but properly document using authenticated package feeds and link to that
3. Put up a password dialog when a 401 occurs
    * This is [what Visual Studio does](https://github.com/bonsai-rx/bonsai/assets/278957/3cf2e041-a58a-4018-a946-3e924e152fd9), but if we go this route I really think we should special-case GitHub Packages since the dialog is confusing in that scenario.
4. Do nothing, hope anyone encountering this problem will come across this issue

[...]

Issue #1804, opened 2024-05-18
Console bootstrapper commands should throw full exception stack trace


glopesdev:
Currently any unhandled exceptions in the package bootstrapper appear to be swallowed and just the error code is returned:
https://github.com/bonsai-rx/bonsai/blob/c5db2c09990d3621b488f89f09acd481ca9b4c5a/Bonsai/Program.cs#L173

While this might make sense in the default UI bootstrapper, it would be useful to have more detailed exception stack traces in the console bootstrapper, so at the very least making a test here and throw if being called with `--no-editor` would help in these situations.

Alternatively we could write to standard error or allow for a verbosity flag.

glopesdev:
Turns out that other exceptions apart from package restore errors would also be swallowed by the current implementation, so we need to take a broader approach here.

PR #1873, opened 2024-07-03
Revamp CI infrastructure and implement release automation

PathogenDavid:
This PR revamps CI and adds release automation.

I don't expect to make any more changes, but I've marked this as a draft because manual configuration to the Bonsai repo must be completed before merging.

As a side effect, this PR brings some other changes as well:
* Bonsai now uses the new [central artifacts folder pattern](https://learn.microsoft.com/en-us/dotnet/core/sdk/artifacts-output), build outputs will be placed in the `artifacts` folder at the root of the repository
  * The stale output directories were removed from the gitignore and will result in a build warning to help people notice and delete them
  * This change includes the previous central `bin` folder 
* Directory.Build.* is now more structured in order to properly support non-C# projects and allow organization of the now more-complex MSBuild infrastructure
* Various cleanup to build system stuff since I was looking at it (removed old cruft, fixed or suppressed various NuGet warnings as appropriate)
* `Bonsai32` no longer clobbers `Bonsai`'s output (this was mostly invisible but was causing chaos within the build could break things in subtle ways)
* Bonsai now builds successfully under the .NET 8 SDK from the command line
* Bonsai now uses modern resource embedding
* Bonsai now *requires* the .NET 8 SDK to be built (enforced via `global.json`) -- This isn't strictly necessary (we do assume .NET 8 since it comes with Source Link) but we're going to have a hard requirement soon so might as well
* ILRepack has been updated to the latest version (required for some modern resource embedding fixes)
* ILRepack no longer happens automatically, it's only done by CI and CI benefits from having it as a separate step anyway
* Introduced an unhealthy amount of Python into a predominately C# project

This PR furthermore makes some major changes to how Bonsai is versioned:

* All packages within Bonsai are versioned together now (packages with no changes skip version numbers instead of getting out of sync)
  * This means Bonsai.Shaders and friends are no longer using initial development (IE: 0.y.z) version numbers
* Some core packages are always released even if they don't have any changes (this list is configurable in `compare-nuget-packages.py`)
* The now-unnecessary revision octet of the file versions has been eliminated (RIP days since first commit Easter egg)
* The origin and purpose of Bonsai binaries is now labeled clearly: Developer (local) builds, unstable CI builds, preview builds, and final release builds
  * Additionally all binaries will be stamped with the Git revision from which they were built (this now shows in the about box as well)
  * The title of Bonsai's editor will be tagged as such for anything other than the final release builds
  * Unstable CI builds from branches other than `main` will have their origin clearly tagged. (For example, a `devel` branch would produce packages versioned like `2.8.5-devel-ci42`)

--------------

Some things that didn't end up making it into this PR:

* The versions of Bonsai packages referenced inside templates (IE: the versions added to generated csproj and such) are not bumped (this had some particularly thorny issues and didn't feel worthwhile)
* This PR does not fix the issue with the bootstrapper restoring the wrong version of itself when it can't find the right one

Related docs repo PR: https://github.com/bonsai-rx/docs/pull/87

Closes https://github.com/bonsai-rx/bonsai/issues/1856
Indirectly fixes https://github.com/bonsai-rx/bonsai/issues/1831 (Bonsai.Editor is always released now, so this scenario can't happen anymore)


Bonsai.Templates/Bonsai.PackageTemplate/Bonsai.PackageTemplate.csproj
```diff
@@ -47,9 +47,6 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Microsoft.VisualStudio.CoreUtility">
-      <Private>False</Private>
-    </Reference>
```
glopesdev, line 52: I don't even remember what this was anymore, I guess happy that it's gone.

Bonsai.Templates/Bonsai.Templates.csproj
```diff
@@ -2,20 +2,31 @@
...
+    <!-- This package only contains .NET templates and does not contain any code to be packed -->
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <IncludeSymbols>false</IncludeSymbols>
+
+    <!-- Ignore warnings about Setup.ps1 as it is not meant to be a NuGet setup script -->
+    <NoWarn>$(NoWarn);NU5110;NU5111</NoWarn>
+
+    <!--
+      This warning basically happens because this package has no assemblies but does have dependencies
+      https://github.com/NuGet/Home/issues/8583
```
glopesdev, line 27:
```suggestion
      This warning basically happens because this package has no assemblies but does have dependencies
      https://github.com/NuGet/Home/issues/8583
```
```suggestion
      This warning happens because this package has no assemblies
      https://github.com/NuGet/Home/issues/8583
```

I guess in this case it doesn't even have dependencies as it is just a template package.

Bonsai.Templates/Bonsai.Templates/source.extension.vsixmanifest
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" ?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
   <Metadata>
-    <Identity Id="BonsaiTemplatesVsix.ac37e8b9-5d6c-407c-b2ca-2e550a8ab918" Version="2.8.2" Language="en-US" Publisher="Bonsai Foundation" />
+    <Identity Id="BonsaiTemplatesVsix.ac37e8b9-5d6c-407c-b2ca-2e550a8ab918" Version="|%CurrentProject%;GetVsixVersion|" Language="en-US" Publisher="Bonsai Foundation" />
```
glopesdev, line 4: Didn't know you could do this, not familiar with this syntax, that's interesting.

Bonsai.Tests.proj
glopesdev, line 1: This was previously needed for CI glad that it's gone.

Bonsai/Bonsai.csproj
```diff
@@ -40,45 +38,84 @@
       <LastGenOutput>Resources.Designer.cs</LastGenOutput>
     </EmbeddedResource>
   </ItemGroup>
-  <Target Name="ILRepack" AfterTargets="Build">
+
+  <!-- Embed System.Resources.Extensions and its dependencies, see Bonsai.SystemResourcesExtensionsSupport -->
```
glopesdev, line 42: Do we still need this even if we are installing it as a dependency later on? I guess you do because of possible resources in NuGet.Design. Curious to know how big that makes the bootstrapper now.

Bonsai/Bonsai.csproj
```diff
@@ -40,45 +38,84 @@
...
-      <InputAssemblies Include="Bonsai.NuGet.Design.dll" />
-      <InputAssemblies Include="Newtonsoft.Json.dll" />
-      <InputAssemblies Include="NuGet.Common.dll" />
-      <InputAssemblies Include="NuGet.Configuration.dll" />
-      <InputAssemblies Include="NuGet.Frameworks.dll" />
-      <InputAssemblies Include="NuGet.Packaging.dll" />
-      <InputAssemblies Include="NuGet.Protocol.dll" />
-      <InputAssemblies Include="NuGet.Resolver.dll" />
-      <InputAssemblies Include="NuGet.Versioning.dll" />
+      <BuiltProjectOutputGroupOutput Remove="@(BuiltProjectOutputGroupOutput)" />
```
glopesdev, line 110: That is a mouthful, can we rename one of the `Output` words with something else like `Files` or `Artifacts`?

Bonsai/Bonsai.csproj
```diff
@@ -40,45 +38,84 @@
...
+      <BuiltProjectOutputGroupOutput Include="$(RepackedOutputFilePath)" />
     </ItemGroup>
+  </Target>
+  <Target Name="UseRepackOutputForSymbolPackage" Condition="'$(UseRepackForBootstrapperPackage)' == 'true'" AfterTargets="DebugSymbolsProjectOutputGroup" DependsOnTargets="ConfigureRepack">
+    <Error Condition="!Exists('$(RepackedOutputSymbolFilePath)')" Text="Project must be repacked before repacking. '$(RepackedOutputFilePath)' not found." />
     <ItemGroup>
-      <InternalizeExcludeAssemblies Include="^Bonsai.Configuration" />
+      <DebugSymbolsProjectOutputGroupOutput Remove="@(DebugSymbolsProjectOutputGroupOutput)" />
```
glopesdev, line 117: Same question here.

Bonsai/Bonsai.csproj
```diff
@@ -40,45 +38,84 @@
...
+
+    <MakeDir Directories="$(RepackedOutputPath)" />
+
     <PropertyGroup>
-      <WorkingDirectory>$(MSBuildThisFileDirectory)bin\$(Configuration)\$(TargetFramework)</WorkingDirectory>
+      <_IlRepackCommand>
+"$(ILRepack)"
+/verbose
+/out:"$(RepackedOutputFilePath)"
+/internalize:"$(MSBuildThisFileDirectory)InternalizeExcludePatterns"
```
glopesdev, line 86: I guess this has to come from a file possibly, so no way to just pass the pattern directly into the command line?

Bonsai/Program.cs
```diff
@@ -41,6 +41,8 @@ static class Program
         [LoaderOptimization(LoaderOptimization.MultiDomainHost)]
         internal static int Main(string[] args)
         {
+            SystemResourcesExtensionsSupport.Initialize();
```
glopesdev, line 44: Do we need to do this for every process, or just for the bootstrapper?

Bonsai/Program.cs
```diff
@@ -170,7 +172,11 @@ internal static int Main(string[] args)
             {
                 var bootstrapper = launchEditor ? (Bootstrapper)new EditorBootstrapper(editorRepositoryPath) : new ConsoleBootstrapper(editorRepositoryPath);
                 try { bootstrapper.RunAsync(Launcher.ProjectFramework, packageConfiguration, editorPath, editorPackageName).Wait(); }
-                catch (AggregateException) { return ErrorExitCode; }
+                catch (AggregateException ex)
+                {
+                    Console.Error.WriteLine(ex);
+                    return ErrorExitCode;
+                }
```
glopesdev, line 179: I somehow feel like this was solved before, will dig through notes for 2.8.3.

Bonsai/SystemResourcesExtensionsSupport.cs
glopesdev, line 1: I wonder if this may have clashes or performance implications with the configuration process later on, which is why I asked above that maybe this is needed only for the pure bootstrapper process.

Bonsai32/Bonsai32.csproj
```diff
@@ -1,21 +1,34 @@
...
+    -->
+    <CopyLocalLockFileAssemblies>false</CopyLocalLockFileAssemblies>
+    <DisableTransitiveProjectReferences>true</DisableTransitiveProjectReferences>
   </PropertyGroup>
   <ItemGroup>
-    <ProjectReference Include="..\Bonsai\Bonsai.csproj" />
+    <ProjectReference Include="..\Bonsai\Bonsai.csproj" Private="false" />
+
+    <!--
+      This reference is necessary to work around a ambiguity in the NuGet restore graph.
```
glopesdev, line 28:
```suggestion
      This reference is necessary to work around an ambiguity in the NuGet restore graph.
```

Interesting, do you have a reference for this? So technically this shouldn't add a dependency to the 32-bit bootstrapper, right? It still only calls the main bootstrapper in 32-bit mode which then resolves everything from its own internal logic.

tooling/Common.csproj.props
```diff
@@ -0,0 +1,39 @@
...
+    <!-- Common C# Properties -->
+    <LangVersion>9.0</LangVersion>
+    <Features>strict</Features>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+
+    <!-- Common NuGet Package Properties -->
+    <Authors>Bonsai Foundation</Authors>
+    <Copyright>Copyright © Bonsai Foundation CIC and Contributors 2011-2024</Copyright>
+    <PackageProjectUrl>https://bonsai-rx.org</PackageProjectUrl>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
```
glopesdev, line 12: Nice, didn't know about `PublishRepositoryUrl`. Just a nitpick for my OCD if we could move this line above so the URLs group together, or even better move the `PublishRepositoryUrl` to the bottom of this section.

tooling/Common.csproj.props
```diff
@@ -0,0 +1,39 @@
...
+    <PackageLicenseUrl>https://bonsai-rx.org/license</PackageLicenseUrl>
+    <PackageIconUrl>https://bonsai-rx.org/assets/images/bonsai.png</PackageIconUrl>
+
+    <!-- Always pack NuGet packages (except on CI where it'll be done separately) -->
+    <GeneratePackageOnBuild Condition="'$(ContinuousIntegrationBuild)' != 'true'">true</GeneratePackageOnBuild>
+
+    <!-- Emit modern symbol pacakges -->
```
glopesdev, line 19:
```suggestion
    <!-- Emit modern symbol packages -->
```

tooling/Common.props
```diff
@@ -0,0 +1,35 @@
...
+    <Platform Condition="'$(Platform)' == ''">AnyCPU</Platform>
+  </PropertyGroup>
+
+  <!-- Build to central artifacts folder instead of scattering bin and obj throughout the solution tree -->
+  <PropertyGroup>
+    <UseArtifactsOutput>true</UseArtifactsOutput>
+    <ArtifactsPath>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)../artifacts'))</ArtifactsPath>
+
+    <!--
+      Odd things can happen when WSL and Windows share the same output directories.
```
glopesdev, line 14: maybe add `, i.e. when building on the same folder using both WSL and Windows`. Or just leave it and I can dig it from this comment.

tooling/Common.targets
glopesdev, line 1: This is detecting the case of `Bonsai.Templates` where two projects share the exact same name.

tooling/Versioning.props
```diff
@@ -0,0 +1,70 @@
...
+    They also can't utilize GetAssemblyVersion to strip off the prerelease suffix becuase it only exists in the modern SDK so we have to do it manually
+  -->
+  <PropertyGroup>
+    <FileVersion>$([System.Text.RegularExpressions.Regex]::Replace('$(Version)', '[\-\+].+$', '')).$(_FileVersionRevision)</FileVersion>
+  </PropertyGroup>
+
+  <!--
+    Reference dummy builds are used to compare NuGet packages to determine what actually changed and needs to be released when releases are published.
+    They work by building and packaging all packages for both the previous version and the next versions of everything and then comparing the results.
+    This means we need to eliminate meaningless differences between versions. We already implicitly use determinsitic builds, so most of this is already taken care of.
```
glopesdev, line 42:
```suggestion
    This means we need to eliminate meaningless differences between versions. We already implicitly use deterministic builds, so most of this is already taken care of.
```

tooling/Versioning.props
```diff
@@ -0,0 +1,70 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <_UnchangedBonsaiVersion>$(Version)</_UnchangedBonsaiVersion>
+    <_UnchangedBonsaiFileVersion>$(FileVersion)</_UnchangedBonsaiFileVersion>
+  </PropertyGroup>
+
+  <Target Name="VersionSanityChecks" BeforeTargets="build">
+    <Error Condition="'$(ContinuousIntegrationBuild)' == 'true' and '$(CiBuildVersion)' == '' and '$(CiBuildVersionSuffix)' == ''" Text="CI version info not configured." />
+    <Error Condition="'$(IsReferenceDummyBuild)' == 'true' and '$(Version)' != '99.99.99'" Text="Reference dummy builds must use hard-coded verison numbers." />
```
glopesdev, line 65:
```suggestion
    <Error Condition="'$(IsReferenceDummyBuild)' == 'true' and '$(Version)' != '99.99.99'" Text="Reference dummy builds must use hard-coded version numbers." />
```

.github/workflows/Bonsai.yml
```diff
@@ -0,0 +1,460 @@
...
+      # ----------------------------------------------------------------------- Get the revision of the main branch
+      # We can't easily bump the version number unless the release happened at main
+      - name: Get the revision of the main branch
+        id: main-revision
+        run: python .github/workflows/gha.py set_output sha `git rev-parse refs/heads/main`
+
+      # ----------------------------------------------------------------------- Bump version number
+      - name: Warn if it doesn't match the release
+        if: steps.main-revision.outputs.sha != github.sha
+        run: python .github/workflows/gha.py print_warning "The main branch is at ${{steps.main-revision.outputs.sha}} but the release was made from ${{github.sha}}, the version number will not be automatically bumped."
```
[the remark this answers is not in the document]
PathogenDavid: Note to self: This might actually be an overly aggressive check

.github/workflows/Bonsai.yml
```diff
@@ -0,0 +1,460 @@
...
+        run: dotnet restore Bonsai.sln
+
+      - name: Build
+        run: dotnet build Bonsai.sln --no-restore --configuration ${{matrix.configuration}}
+
+      # ----------------------------------------------------------------------- Repack bootstrapper
+      # This happens before pack since the bootstrapper package uses it
+      - name: Repack bootstrapper
+        if: matrix.create-installer || env.UseRepackForBootstrapperPackage == 'true'
+        run: dotnet build Bonsai --no-restore --configuration ${{matrix.configuration}} -t:Repack
```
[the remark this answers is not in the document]
PathogenDavid: Correct. The only reason the Bonsai.sln is repeated everywhere is that `dotnet` sees the `Bonsai.Setup.sln` and doesn't want to assume which solution you care about. (Ideally we'd just omit the solution from all commands otherwise.)

.github/workflows/Bonsai.yml
```diff
@@ -0,0 +1,460 @@
...
+
+      - name: Build Setup
+        id: create-installer
+        if: matrix.create-installer
+        run: msbuild /nologo /maxCpuCount Bonsai.Setup.sln /p:Configuration=${{matrix.configuration}}
+      
+      # ----------------------------------------------------------------------- Collect artifacts
+      - name: Collect NuGet packages
+        uses: actions/upload-artifact@v4
+        if: matrix.collect-packages && steps.pack.outcome == 'success' && always()
```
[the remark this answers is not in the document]
PathogenDavid: The main reason for it is that it allows this step to run even if previous steps failed.

The main idea is that if a build is partially failing for some reason that you still get access to whatever artifacts did manage to build before the failure, which is frequently useful for debugging and is helpful to someone who wants some of the artifacts.

For example: If building the Visual Studio templates failed for some reason, you'd still have access to the portable zip.

To frame it another way: `Collecting NuGet Packages` really only cares that `collect-packages` was successful, so it will always run as long as it's successful (even if other things failed.)

Under the hood, every single `if` in GitHub Actions has an implicit `success() &&` prefix. Adding [things like `always()`](https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions) anywhere to the expression is disabling that behavior.

.github/workflows/Bonsai.yml
```diff
@@ -0,0 +1,460 @@
...
+          if-no-files-found: error
+          path: artifacts/package/${{matrix.configuration-lower}}/**
+
+      - name: Collect portable zip
+        uses: actions/upload-artifact@v4
+        if: steps.create-portable-zip.outcome == 'success' && always()
+        with:
+          name: PortableZip${{matrix.artifacts-suffix}}
+          if-no-files-found: error
+          path: artifacts/Bonsai.zip
```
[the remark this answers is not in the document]
PathogenDavid: Not that I'm aware of. It's an ambiguous release artifact that has no relationship with the rest of the Bonsai build system, it could go anywhere really.

A lot of CI pipelines just emit this sort of stuff into the repo root since it'll evaporate into oblivion as soon as the CI run ends (but that feels a bit uncouth to me.)

.github/workflows/Bonsai.yml
```diff
@@ -0,0 +1,460 @@
...
+        run: python .github/workflows/compare-nuget-packages.py artifacts/Packages-dummy-prev/ artifacts/Packages-dummy-next/ artifacts/Packages/ artifacts/ReleaseManifest
+
+      # ----------------------------------------------------------------------- Collect release manifest
+      - name: Collect release manifest
+        uses: actions/upload-artifact@v4
+        if: steps.compare-packages.outcome != 'skipped' && always()
+        with:
+          name: ReleaseManifest
+          if-no-files-found: error
+          path: artifacts/ReleaseManifest
```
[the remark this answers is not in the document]
PathogenDavid: I would worry that shuttling these into the same folder like that might imply they're related somehow (which they really aren't) or they're created by the same tool (which again they aren't.)

Maybe a good way to look at these files is that they're build artifacts that defy categorization.

These files really only ever exist on a filesystem that humans never really interact with.

When someone does interact with these files (in the context of a filesystem), it's because they're testing the CI scripts locally in which case it's just gonna use whatever location they specify.

.github/workflows/Bonsai.yml
```diff
@@ -0,0 +1,460 @@
...
+          name: Installer
+
+      # ----------------------------------------------------------------------- Filter NuGet packages
+      - name: Filter NuGet packages
+        run: python .github/workflows/filter-release-packages.py ReleaseManifest Packages
+
+      # ----------------------------------------------------------------------- Upload release assets
+      - name: Upload release assets
+        if: github.event_name == 'release'
+        run: gh release upload ${{github.event.release.tag_name}} Bonsai.zip Bonsai-*.exe --clobber
```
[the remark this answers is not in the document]
PathogenDavid: Haha, I found that amusing too. Much more fun than `--force` (less ambiguous too, imo)

.github/workflows/Bonsai.yml
```diff
@@ -0,0 +1,460 @@
...
+        env:
+          GH_TOKEN: ${{github.token}}
+
+      # ----------------------------------------------------------------------- Push to GitHub Packages
+      - name: Push to GitHub Packages
+        run: dotnet nuget push "Packages/*.nupkg" --skip-duplicate --no-symbols --api-key ${{secrets.GITHUB_TOKEN}} --source https://nuget.pkg.github.com/${{github.repository_owner}}
+        env:
+          # This is a workaround for https://github.com/NuGet/Home/issues/9775
+          DOTNET_SYSTEM_NET_HTTP_USESOCKETSHTTPHANDLER: 0
```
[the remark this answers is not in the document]
PathogenDavid: We talked about this during our meeting, but for the sake of posterity:

I've always specified `--api-key` here and still had this issue in the past.

The issue this is working around manifests itself as random intermittent failure of pushing packages, which is extraordinarily frustrating since it happens so late in the workflow.

In retrospect I should've taken it out when I was developing the workflow to see if it was truly necessary anymore, but I'd lean towards just leaving it.

.github/workflows/Bonsai.yml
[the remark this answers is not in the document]
PathogenDavid: We talked about this during our meeting, but I looked at it more now and I remembered why this exists.

This part of the workflow pushes commits, those commits have to go to some branch, we choose `main` since that's sensible.

However releases are made from tags, not branches. If someone were to make a release from another branch (EG: someone backporting a critical fix to Bonsai 2.9 after Bonsai 3.0 has released) we wouldn't want to commit the version bump to `main` since the release didn't happen from there.

The fact that this also skips version bumping when someone pushes commits to `main` during the workflow run isn't really intentional. There's no really a straightforward way to avoid that edge case though, and I don't think it's really worth handling. (If the Bonsai repo is ever so busy that this becomes a concern this automated version bumping will likely need to have been changed anyway.)

I will reword some comments to clarify that this is what the intent is here.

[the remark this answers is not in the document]
PathogenDavid: Also it's worth noting that bumping the version number isn't super critical. It only impacts the version used for dev builds and unstable builds. For actual releases (and pre-releases) the version specified on the GitHub interface is always what's used.

(This is also why it doesn't really matter if the automatic version isn't a version we actually intend to release. For example it'll be 2.8.5 after we release 2.8.4 even though we probably won't actually release a 2.8.5)

.github/workflows/compare-nuget-packages.py
```diff
@@ -0,0 +1,223 @@
...
+    
+    # The package doesn't exist at all, assume mistake unless we're checking the optional symbol packages
+    if not a_path.exists():
+        if is_snupkg:
+            verbose_log("Equivalent: Neither package exists")
+            return True
+        raise FileNotFoundError(f"Neither package exists: '{a_path}' or '{b_path}'")
+    
+    # From this point on: Check everything and emit messages for debugging purposes
+    is_equvalent = True
```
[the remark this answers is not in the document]
PathogenDavid: The downside of Intellisense 😔

.github/workflows/create-portable-zip.py
```diff
@@ -0,0 +1,59 @@
...
+import gha
+
+if len(sys.argv) != 3:
+    gha.print_error('Usage: create-portable-zip.py <output_path> <release|debug>')
+    sys.exit(1)
+else:
+    output_path = sys.argv[1]
+    configuration = sys.argv[2].lower()
+
+with zipfile.ZipFile(output_path, 'x', zipfile.ZIP_DEFLATED, compresslevel=9) as output:
```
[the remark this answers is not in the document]
PathogenDavid: I said during our meeting that compresslevel here was to override the default of not compressing at all. That wasn't right, that is true for the previous parameter (specified as ZIP_DEFLATED) here.

I specified it as 9 mainly because the Python documentation didn't specify what the default was otherwise and this step is so fast that speed wasn't a concern.

I dug around in the Python source code and found that omitting it passes `Z_DEFAULT_COMPRESSION` to zlib which [documents that as level 6](http://www.zlib.net/manual.html).

--------

I was also curious about the savings so I did a little experiment and it's actually a bit bigger lol.

* Bonsai.zip as published on GitHub: 1.34 MB
* Bonsai.zip compressed by Python at level 9: 1.38 MB
* Bonsai.zip compressed by 7zip at level 9: 1.32 MB

.github/workflows/create-portable-zip.py
```diff
@@ -0,0 +1,59 @@
...
+    output.mkdir('Extensions')
+    output.mkdir('Gallery')
+
+    output.write(f'artifacts/bin/Bonsai/{configuration}-repacked/Bonsai.exe', 'Bonsai.exe')
+    output.write(f'artifacts/bin/Bonsai/{configuration}/Bonsai32.exe', 'Bonsai32.exe')
+
+    nuget_config = [
```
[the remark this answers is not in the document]
PathogenDavid: As mentioned during our meeting I hadn't noticed that file when I was working on this aspect. (Portable zip was done way before the installer.)

Thinking about this more, I wonder if the opposite might be the better strategy. Delete `Bonsai.Setup/Bonsai.config` and have the shipped NuGet.config come from Python.

I actually considered doing this in order to allow the installer to have a more appropriate NuGet.config corresponding to the CI job, but I didn't see most people wanting to use installers from unstable releases or testing forks. (Especially not until https://github.com/bonsai-rx/bonsai/issues/1849 happens since the `Bonsai.config` is hard to discover with the installer so nobody would see the instructions I wrote 😅)

.github/workflows/create-portable-zip.py
```diff
@@ -0,0 +1,59 @@
...
+        '    <add key="Bonsai Packages" value="https://www.myget.org/F/bonsai/api/v3/index.json" />',
+        '    <add key="Community Packages" value="https://www.myget.org/F/bonsai-community/api/v3/index.json" />',
+    ]
+
+    nuget_api_url = os.getenv('NUGET_API_URL')
+    if nuget_api_url is not None:
+        nuget_config.append(f'    <add key="NuGet Package Testing Feed" value="{nuget_api_url}" />')
+
+    # Unstable builds of Bonsai will automatically reference the GitHub Packages feed
+    if os.getenv('IS_FULL_RELEASE') == 'false':
```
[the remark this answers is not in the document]
PathogenDavid: As mentioned during our meeting, it's not really that much (it just looks like a lot since I included instructions to work around https://github.com/bonsai-rx/bonsai/issues/1849)

More important that the GitHub Packages aspect though is that "NuGet Package Testing Feed". That feature is invaluable for testing this workflow in forks and I'd like to keep it.

.github/workflows/nuget.py
```diff
@@ -0,0 +1,62 @@
...
+
+def is_preview_version(version: str) -> bool:
+    match = package_version_regex.match(version)
+    if match is None:
+        gha.print_error(f"Version '{version}' is not a legal semver version string!")
+        return True
+    
+    return match.group('prerelease') is not None
+
+class SemverVersion:
```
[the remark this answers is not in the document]
PathogenDavid: Haha yeah, that's a weird name. Went with `SemanticVersion`

.gitignore
```diff
@@ -1,10 +1,7 @@
...
+/artifacts/
+/packages/
 Resources/*.ico
 *.Generated.cs
-*.orig
 *.user
-*.suo
-*.exe
-.vs
+Bonsai.Setup.Bootstrapper/vc_redist.*.exe
```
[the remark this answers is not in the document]
PathogenDavid: We talked about this in our meeting, but for posterity:

.exe: I strongly prefer narrowly defined `.gitignore`s since they help you notice when a mistake is made. (Executables shouldn't be in Bonsai's source tree outside of this one case, so if one appears it might indicate something is messed up in the build system or something.)

`.suo`: Yeah, the loose `suo` hasn't been a thing since like Visual Studio 2013 I think?

Bonsai.Design/Bonsai.Design.csproj
```diff
@@ -1,15 +1,14 @@
...
     <Title>Bonsai - Design Library</Title>
     <Description>Bonsai Design Library containing base visualizer classes and editor infrastructure.</Description>
     <PackageTags>Bonsai Design Rx Reactive Extensions</PackageTags>
     <UseWindowsForms>true</UseWindowsForms>
     <TargetFramework>net462</TargetFramework>
-    <VersionPrefix>2.8.1</VersionPrefix>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="jacobslusser.ScintillaNET" Version="3.6.3" />
+    <PackageReference Include="System.Resources.Extensions" Version="8.0.0" Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'" />
```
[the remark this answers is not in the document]
PathogenDavid: It was already there for some other core dependencies but yeah 😅

Bonsai.Editor/AboutBox.Designer.cs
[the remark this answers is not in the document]
PathogenDavid: I briefly mentioned it during our meeting, but this one makes a lot more sense if you read the commit message:

https://github.com/bonsai-rx/bonsai/pull/1873/commits/c274a4f11a14c08850d8c08fafa4761324474f5c

Basically enabling Source Link made the version string longer which made it invisible in the about box. This change is primarily to fix that.

Bonsai.Editor/AboutBox.Designer.cs
```diff
@@ -70,89 +80,81 @@ private void InitializeComponent()
...
             this.textBoxDescription.Dock = System.Windows.Forms.DockStyle.Fill;
-            this.textBoxDescription.Location = new System.Drawing.Point(143, 107);
+            this.textBoxDescription.Location = new System.Drawing.Point(153, 107);
             this.textBoxDescription.Margin = new System.Windows.Forms.Padding(6, 3, 3, 3);
-            this.textBoxDescription.Multiline = true;
             this.textBoxDescription.Name = "textBoxDescription";
             this.textBoxDescription.ReadOnly = true;
-            this.textBoxDescription.ScrollBars = System.Windows.Forms.RichTextBoxScrollBars.Both;
-            this.textBoxDescription.Size = new System.Drawing.Size(271, 126);
+            this.textBoxDescription.Size = new System.Drawing.Size(309, 126);
```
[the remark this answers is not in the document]
PathogenDavid: Yeah this diff is messier than I'd like but I've learned better than to defy the whims of the form designer.

All the changes are intentional though. See previous comment.

Bonsai.Editor/AboutBox.Designer.cs
```diff
@@ -70,89 +80,81 @@ private void InitializeComponent()
...
             this.labelCopyright.Name = "labelCopyright";
-            this.labelCopyright.Size = new System.Drawing.Size(271, 17);
+            this.labelCopyright.Size = new System.Drawing.Size(309, 17);
             this.labelCopyright.TabIndex = 21;
             this.labelCopyright.Text = "Copyright";
             this.labelCopyright.TextAlign = System.Drawing.ContentAlignment.MiddleLeft;
             // 
             // labelCompanyName
             // 
+            this.labelCompanyName.AutoEllipsis = true;
```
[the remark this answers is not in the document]
PathogenDavid: See my comment on the main feedback for this file.

Bonsai.Editor/AboutBox.cs
```diff
@@ -11,14 +11,32 @@ partial class AboutBox : Form
...
+            this.Text = $"About {AssemblyTitle}{BuildKindTitleSuffix}";
             this.labelProductName.Text = AssemblyProduct;
-            this.labelVersion.Text = String.Format("Version {0}", AssemblyVersion);
+            this.labelVersion.Text = $"Version {AssemblyVersion}";
             this.labelCopyright.Text = AssemblyCopyright;
             this.labelCompanyName.Text = AssemblyCompany;
             this.textBoxDescription.Text = AssemblyDescription + Environment.NewLine + Resources.AttributionNotices;
         }
 
+        internal static string BuildKindTitleSuffix
```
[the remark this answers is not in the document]
PathogenDavid: As mentioned during our meeting this is a somewhat intentionally clunky name to makes it clear that it has a very narrow use-case and isn't suitable for anything other than being a form title suffix due to the spaces included in most (but not all) of the branches.

Bonsai.Editor/EditorForm.cs
```diff
@@ -187,6 +187,8 @@ public partial class EditorForm : Form
                 editorControl.Leave += delegate { menuStrip.Enabled = true; };
             }
             components.Add(editorControl);
+
+            UpdateTitle();
```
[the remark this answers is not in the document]
PathogenDavid: I think here is fine, the point of the constructor on forms is the set up the visual aspects of the form before it's displayed and this is part of that.

Bonsai.Templates/Bonsai.PackageTemplate/Bonsai.PackageTemplate.csproj
PathogenDavid: Yeah I couldn't find a straight answer on what this was either, but it was causing problems and didn't seem to be being used so off it went.

Bonsai.Templates/Bonsai.Templates.csproj
PathogenDavid: Good catch, but I left the word "basically" since the warning doesn't literally happen because there's no assembly, it's just the first domino which triggers a chain reaction which causes that warning.

When you encounter the warning in the wild you'd never guess it was due to there being no assembly in the package:

> warning NU5128: Some target frameworks declared in the dependencies group of the nuspec and the lib/ref folder do not have exact matches in the other location.

Bonsai.Tests.proj
PathogenDavid: Yup, the modern way of just running `dotnet test` is so much nicer

Bonsai.Templates/Bonsai.Templates/source.extension.vsixmanifest
PathogenDavid: Yup! I didn't know about it until this PR either. It basically invokes this MSBuild target to get the version it should use:

https://github.com/bonsai-rx/bonsai/blob/93503f2a059eda7381ed20d74d69fb0923a8ae02/Bonsai.Templates/Bonsai.Templates/Bonsai.Templates.csproj#L106-L108

I discovered it by looking at the Roslyn source code ([example](https://github.com/dotnet/roslyn/blob/967907da1e4846c15b30a68e37ddcdfc9964781a/src/Compilers/Extension/source.extension.vsixmanifest#L4)) since I knew they deploy a lot of VSIX and figured there was no way they were manually bumping versions.

During our meeting I incorrectly stated that this target is normally something in the modern .NET SDK. It actually turns out it's [part of .NET Arcade](https://github.com/dotnet/arcade/blob/4a7d983f833d6b86365ea1b2b4d6ee72fbdbf944/src/Microsoft.DotNet.Arcade.Sdk/tools/VisualStudio.targets#L31-L32) This made me realize that the SDK-style VSIX projects used by Roslyn probably come from there too, which would definitely explain why I couldn't find them documented anywhere 😅

Bonsai/Bonsai.csproj
PathogenDavid: As noted during our meeting, yes this unfortunately has to be part of the bootstrapper. It can't show the package installation dialog without it.

The compressed size of the `Bonsai.exe` increased 0.06 MB as a result of this PR. These backport assemblies are not very big. (`System.Memory.dll` is only 139 KB uncompressed.)

Bonsai/Bonsai.csproj
PathogenDavid: As noted during our meeting, I didn't name these. The .NET SDK devs did 😅

Bonsai/Bonsai.csproj
PathogenDavid: Same response here.

Bonsai/Bonsai.csproj
PathogenDavid: Unfortunately not. I even dug into the ILRepack source code to double check that I wasn't just misunderstanding something. IIRC the old ILRepack build task we were using was emitting a temporary file to disk for this purpose.

Bonsai/Program.cs
PathogenDavid: I tried moving it to be bootstrapper-only and it goes bang trying to resolve `System.Runtime.CompilerServices.Unsafe` while loading the start screen. (The failed assembly load was initiated by `Bonsai.Configuration`, so it at least *tried*.)

As noted during our meeting I'm very uncomfortable with the idea of only doing this sometimes.

The dependencies of `System.Resources.Extensions` are all EOL and will not be updated so we will not need to handle the case of someone installing a package which wants a newer version.

`System.Resources.Extensions` itself still receives new versions, but I'm not very concerned about people wanting a different version from Bonsai. If someone does I will personally fly to their house and throw an plastic bonsai tree at them.

If I never have to think about ILRepacking this package again it will be too soon.

Bonsai/Program.cs
PathogenDavid: I believe you're thinking of https://github.com/bonsai-rx/bonsai/pull/1805, which was a slightly different scenario for a different exception path.

Bonsai/SystemResourcesExtensionsSupport.cs
PathogenDavid: Clashes: As noted earlier I find it unlikely, but if it makes you feel better we could reinstall the resolver after `ConfigurationHelper.SetAssemblyResolve` to ensure it runs afterwards.

Performance: Negligible, the amount of time spent trying to load the non-existent resource would be easily dwarfed by the overhead of loading the assembly itself, but they say the proof is in the BenchmarkDotNet pudding:

| Method                 | Mean     | Error     | StdDev    |
|----------------------- |---------:|----------:|----------:|
| UselessAssemblyResolve | 3.022 us | 0.0155 us | 0.0121 us |

<details>
<summary>Benchmark Source</summary>

```csharp
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using System.Diagnostics;
using System.Reflection;

BenchmarkRunner.Run<Benchmark>();

public class Benchmark
{
    public string Name = "System.CompilerServices.Unsafe";

    [Benchmark]
    public Assembly? UselessAssemblyResolve()
    {
        var assemblyName = new AssemblyName(Name);
        using var embeddedAssembly = typeof(Benchmark).Assembly.GetManifestResourceStream($"{"Bonsai"}.{assemblyName.Name}.dll");

        if (embeddedAssembly is null)
            return null;

        var assemblyBytes = new byte[embeddedAssembly.Length];
        int readLength = embeddedAssembly.Read(assemblyBytes, 0, assemblyBytes.Length);
        Debug.Assert(readLength == assemblyBytes.Length);

        var result = Assembly.Load(assemblyBytes);
        Debug.WriteLine($"Redirecting '{Name}' to embedded '{result.FullName}'");
        return result;
    }
}
```

(This does exclude some minor things like the delegate invocation and allocating the extra `ResolveEventArgs` but those are known to be negligible in this situation.)

</details>

Bonsai32/Bonsai32.csproj
PathogenDavid: Yeah, this is a pure restore-time issue.

No reference, this is a very specific weird case caused by how Bonsai is structured.

I don't remember all the details (this is like one of the first things I did on this PR), but basically Bonsai32 can "see" an indirect dependency to `System.Runtime.CompilerServices.Unsafe/4.5.3` in the restore graph via the path `Bonsai32` -> `Bonsai` -> `Bonsai.Design` -> `System.Resources.Extensions/8.0.0` -> `System.Memory/4.5.5` -> `System.Runtime.CompilerServices.Unsafe/4.5.3`

Bonsai32 restore graph visualization

Meanwhile there exists within Bonsai an indirect dependency to `System.Runtime.CompilerServices.Unsafe/6.0.0` via the path `Bonsai` -> `System.Reflection.MetadataLoadContext/7.0.0` -> `System.Collections.Immutable/7.0.0` -> `System.Runtime.CompilerServices.Unsafe/6.0.0` as well as the similar path to `System.Runtime.CompilerServices.Unsafe/4.5.3` through `System.Memory/4.5.5` as seen by Bonsai32:

Bonsai restore graph visualization

In the case of Bonsai it resolves to 6.0.0 since it's the newest dependency and satisfies every package's constraints.

However Bonsai32 does not see the indirect 6.0.0 dependency in its restore graph because Bonsai marks `System.Reflection.MetadataLoadContext` as private.

NuGet doesn't handle this situation properly (I think it catches it too late) and you get MSB3277 as a result. (The comment is maybe a bit wordy since this is unlikely to happen again, but my hope was I could save someone some time in case it somehow did.)

tooling/Common.csproj.props
PathogenDavid: Yup, this is technically part of Source Link.

Haha yeah that was a little gross of me, it's at the end so it's not splitting the URLs anymore.

tooling/Common.csproj.props
PathogenDavid: Somehow this typo doesn't exist in Biohazrd, must've copy+pasted that one manually 😅

https://github.com/MochiLibraries/Biohazrd/blob/4dbaabfb8195853a931a21d9b6468baed9020bf0/tooling/Common.csproj.props#L22

.github/workflows/Bonsai.yml
```diff
@@ -0,0 +1,460 @@
...
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git commit -m "Bump to version \`$NEXT_VERSION\`" --allow-empty
+
+      - name: Push changes
+        if: steps.main-revision.outputs.sha == github.sha
+        run: git push
+
+      # ----------------------------------------------------------------------- Close milestone
+      - name: Close milestone
+        uses: actions/github-script@v7
```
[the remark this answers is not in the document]
PathogenDavid: Yeah! It's really nice for filling in the gaps of the GitHub CLI or doing things that need a bit more code.

.github/workflows/Bonsai.yml
```diff
@@ -0,0 +1,460 @@
...
+
+      # ----------------------------------------------------------------------- Push to NuGet.org
+      - name: Push to NuGet.org
+        run: dotnet nuget push "Packages/*.nupkg" --api-key ${{secrets.NUGET_API_KEY}} --source ${{vars.NUGET_API_URL}}
+        env:
+          # This is a workaround for https://github.com/NuGet/Home/issues/9775
+          DOTNET_SYSTEM_NET_HTTP_USESOCKETSHTTPHANDLER: 0
+
+      # ----------------------------------------------------------------------- Dispatch docs repo update
+      # This might seem like an odd spot to do this, but we need access to the PublicRelease environment for our secrets and using it in two jobs means two approvals are needed
```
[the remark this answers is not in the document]
PathogenDavid: We talked about this in our meeting, but for the sake of posterity:

This separation is also important for ensuring that the `PublicRelease` environment is only referenced in proper release workflow runs.

Additionally I think it's valuable to have the GitHub job act as a dry run of pushing to NuGet.org.

.github/workflows/create-portable-zip.py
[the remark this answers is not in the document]
PathogenDavid: I was just going to go ahead and do this but it's getting late. I think it can wait for a separate PR if you like the idea. (I agree that messing with the XML DOM is a pain. We could make it a substitution template similar to what I did in the docs repo, but then we'd have this ugly substitution marker in our release NuGet.config files.)

tooling/Common.targets
PathogenDavid: I believe you intended to leave this comment here:

https://github.com/bonsai-rx/bonsai/blob/399f50fdb9b006fa2ead0d7ec428749f64c5ecf7/tooling/Common.props#L24-L25

I ended up adding some spacing here so that it's less wall-of-texty, so hopefully that helps it stand out that this is a special case.

Bonsai/Program.cs
glopesdev: Ah, that's right, and that reminds me why I think I left this here: if you run the bootstrapper from the command line, it will print the error twice, once from the console logger and again from the Console.Error here. I will try to verify this is the case.

glopesdev: @PathogenDavid actually this is indeed indirectly related to #1805, which is a proposed fix to #1804 which discusses exactly this issue. #1805 was supposed to fix it since any package bootstrap errors will be routed to either a UI or the console logger standard error anyway.

Most likely what is happening here if you are hitting this are errors outside the NuGet client workflow, I would guess for example the assembly resolution and loading errors? Those would crash the process silently, which I agree is not great.

The best might be to revert #1805 by removing the `throw` statement after logging to make the behavior identical to the UI bootstrapper (i.e. if the exception is logged, it is not rethrown). In that case printing any other unhandled exceptions in the catch statement as you did at the top-level should not generate duplicate error traces.

glopesdev: The above is also not entirely satisfactory since not throwing in the console bootstrapper will prevent returning a proper error exit code which can be useful for CLI evaluation. Maybe we just merge this anyway and reopen #1804 and I can submit a better PR to address this later.

glopesdev (approved):

.github/workflows/create-portable-zip.py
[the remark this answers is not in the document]
PathogenDavid: Created issue https://github.com/bonsai-rx/bonsai/issues/1881

PR #1876, opened 2024-07-03
Ensure Dequeue is not attempted if buffer size is 0

### LINKED ISSUE #1875: `ObjectTextVisualizer` attempts to Dequeue empty buffer
To reproduce:

Use the following workflow and layout file. Ensure the tabcontrol visualizer is active at startup.
[src.zip](https://github.com/user-attachments/files/16090655/src.zip)

After attaching the process to VS it seems the error is thrown by this line:

https://github.com/bonsai-rx/bonsai/blob/82448b97a414291b71007e0b564f031ad91b7e8e/Bonsai.Design/ObjectTextVisualizer.cs#L40-L41

I believe that due to the small size of the window, the following line returns 0:
`bufferSize = (int)((textBox.ClientSize.Height - 2) / lineHeight);`

and the `while` condition evaluates to true. 

The fix should be to change the while condition to also check if the size is not 0

bruno-f-cruz:
Closes #1875

### DIFF
--- a/Bonsai.Design/ObjectTextVisualizer.cs
+++ b/Bonsai.Design/ObjectTextVisualizer.cs
@@ -37,7 +37,7 @@ namespace Bonsai.Design
             var sb = new StringBuilder();
 
             // Trim old values if bufferSize was reduced
-            while (buffer.Count >= bufferSize)
+            while ((buffer.Count >= bufferSize) & (buffer.Count > 0))
                 buffer.Dequeue();
 
             // Add new values to the buffer (and only the ones which might appear)


Bonsai.Design/ObjectTextVisualizer.cs
```diff
@@ -37,7 +37,7 @@ protected override void ShowBuffer(IList<Timestamped<object>> values)
             var sb = new StringBuilder();
 
             // Trim old values if bufferSize was reduced
-            while (buffer.Count >= bufferSize)
+            while ((buffer.Count >= bufferSize) & (buffer.Count > 0))
```
PathogenDavid, line 40: Should be `&&`, which would also make these groupings unnecessary

PathogenDavid (changes requested):
My original intent here was that `bufferSize` should always be greater than 0, I just didn't enforce it robustly enough.

A better approach I think would be to just enforce that, IE: Replace

https://github.com/bonsai-rx/bonsai/blob/a9dd431b6b84ad60620ec5244183a130fbe2dd3f/Bonsai.Design/ObjectTextVisualizer.cs#L167

with

```csharp
bufferSize = Math.Max(1, (int)((textBox.ClientSize.Height - 2) / lineHeight));
```

That way if the buffer size would've been 0, there's always at least a single entry kept around so it appears when the user resizes the form.

(As it's written right now there's also other places assuming `bufferSize` is non-zero too so they'd still be broken too.)

bruno-f-cruz:
Sounds good! Feel free to edit the PR directly or just open a new one! Thanks for looking into it!

PathogenDavid:
Created https://github.com/bonsai-rx/bonsai/pull/1882 (there was one other small change that was needed so it was a good thing I took a closer look.)

PR #1882, opened 2024-07-04
Fix `ObjectTextVisualizer` allowing a buffer size of 0

PathogenDavid:
There also existed an edge case where `bufferSize` was 0 if the visualizer never got painted.

Fixes https://github.com/bonsai-rx/bonsai/issues/1875
Supersedes https://github.com/bonsai-rx/bonsai/pull/1876

### DIFF
--- a/Bonsai.Design/ObjectTextVisualizer.cs
+++ b/Bonsai.Design/ObjectTextVisualizer.cs
@@ -23,7 +23,7 @@ namespace Bonsai.Design
         RichTextBox textBox;
         UserControl textPanel;
         Queue<string> buffer;
-        int bufferSize;
+        int bufferSize = 1;
 
         /// <inheritdoc/>
         protected override int TargetInterval => 1000 / 30;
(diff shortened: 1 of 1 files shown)


Bonsai.Design/ObjectTextVisualizer.cs
glopesdev, line 1: The only thought which came to mind is the use of markdown in the commit message. I have so far preferred to type everything in plain text, so just wondering what should be the guideline here. A brief glance at other popular repos was not entirely decisive although it does seem that markdown usage is always a minority, with most references to type names in plain text rather than backticked.

glopesdev: Also just realized that in some contexts [backtick can signify command execution](https://stackoverflow.com/questions/71155954/backticks-in-git-commit-message), which is not something that had occurred to me previously and makes me lean a bit towards avoiding their use at least in commit message titles.

PathogenDavid: GitHub (and other software forges) render it as monospace, even when it's not rendered in monospace I think it's more comfortable to read and clarifies it's referring to a concept from code.

> some contexts

In `sh` and its derivatives, sure. I don't really see why this is relevant, the linked question doesn't actually have anything to do with Git commit messages. It's a confused newbie using their terminal incorrectly.

I don't write my commit messages on the terminal in the first place. Anyone who does should've learned this already or learn to do so quickly like the newbie you linked to.

I don't think we need to strive for literally everything in Bonsai to be homogeneous. This feels excessively nitpicky.

I also don't really think popular repos doing a thing is a super great indicator of whether or not it should be done. If we followed the crowd on everything we wouldn't write meaningful commit messages in the first place and Bonsai would be an Electron app powered by a fleet of overpriced AWS garbage.

glopesdev: Fair enough, I buy the argument and agree it's better not to too overboard on this as long as the information is there.

glopesdev (approved):

Issue #1881, opened 2024-07-04
Generate Bonsai.Setup's Bonsai.config from the same source as the portable zip


PathogenDavid:
As discussed here https://github.com/bonsai-rx/bonsai/pull/1873#discussion_r1665158852

Right now the NuGet config for the installer and the portable zip comes from two separate places. It should come from a single source of truth.

Since Bonsai.Setup will mostly only ever be built by CI now, the easiest route would be to just generate it from the same Python script as `create-portable-zip.py`, or maybe use MSBuild instead to avoid the Python dependency on building Bonsai.Setup.

PR #1887, opened 2024-07-05
Make package comparison in CI optional for the sake of forks

PathogenDavid:
Minor change to release automation I realized should be done after I saw your fork as failing CI.

Forks won't always have the `latest` tag (and even if they do it might be stale), so don't bother with package comparison unless the fork explicitly enables it.

⚠ Before merging you need to [add an Actions variable](https://github.com/bonsai-rx/bonsai/settings/variables/actions) `ENABLE_PACKAGE_COMPARISON` = `true` to this repo.

(Forks can also choose to define that variable to assert that their `latest` tag is valid and that they want the comparisons done.)

### DIFF
--- a/.github/workflows/create-build-matrix.py
+++ b/.github/workflows/create-build-matrix.py
@@ -43,9 +43,18 @@ def add_dummy(name: str, artifacts_suffix: str):
     dummy['artifacts-suffix'] = artifacts_suffix
     return dummy
 
-if os.getenv('GITHUB_EVENT_NAME') != 'pull_request':
+enable_package_comparison = os.getenv('enable_package_comparison') == 'true'
+github_event_name = os.getenv('GITHUB_EVENT_NAME')
+
+if github_event_name != 'pull_request' and enable_package_comparison:
     add_dummy('Previous Dummy', '-dummy-prev')['checkout-ref'] = 'refs/tags/latest'
     add_dummy('Next Dummy', '-dummy-next')
+
+# Fail early if we won't be able to do package comparison and the run must publish packages to make logical sense
+# Package comparison requires the `latest` tag to exist, but it will usually either be missing or invalid for forks so we require it to be opt-in
+if not enable_package_comparison:
+    if github_event_name == 'release' or (github_event_name == 'workflow_dispatch' and os.getenv('will_publish_packages') == 'true'):
+        gha.print_error('Release aborted. We would not be able to determine which packages need to be released as this repository is not configured for package comparison.')
 
 # Output
 matrix_json = json.dumps({ "include": matrix }, indent=2)
--- a/.github/workflows/Bonsai.yml
+++ b/.github/workflows/Bonsai.yml
@@ -45,6 +45,9 @@ jobs:
       - name: Create build matrix
         id: create-matrix
         run: python .github/workflows/create-build-matrix.py
+        env:
+          enable_package_comparison: ${{vars.ENABLE_PACKAGE_COMPARISON}}
+          will_publish_packages: ${{github.event.inputs.will_publish_packages}}
 
   # =====================================================================================================================================================================
   # Build, test, and package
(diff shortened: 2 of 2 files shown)


.github/workflows/create-build-matrix.py
```diff
@@ -43,10 +43,19 @@ def add_dummy(name: str, artifacts_suffix: str):
     dummy['artifacts-suffix'] = artifacts_suffix
     return dummy
 
-if os.getenv('GITHUB_EVENT_NAME') != 'pull_request':
+enable_package_comparison = os.getenv('enable_package_comparison') == 'true'
+github_event_name = os.getenv('GITHUB_EVENT_NAME')
```
glopesdev, line 47: Is `os.getenv` case-insensitive? The reason I was wondering is I added the repository variable `ENABLE_PACKAGE_COMPARISON` as upper snake case but then noticed here is lower snake case.

I can see that in the build workflow there is a lower snake case field `enable_package_comparison` definition in the `env` section, which probably is what is used here? However, I was still then left wondering about `GITHUB_EVENT_NAME` which is in upper snake case in the python script, but lower snake case in the `env` section and I couldn't find it defined elsewhere.

Mostly just a clarification to see if I am understanding and following the logic of the action environment correctly.

PathogenDavid: You caused me to think way to critically about my GitHub Actions naming convention habits and I rambled a bit 😅

Short version:

* `os.getenv` case sensitivity is OS-dependent
* Yes it comes from the `env` block and that's why it's lowercase
* `GITHUB_EVENT_NAME` is ambient state, [it's documented here](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)
  * I could switch it to be passed explicitly like `configure-build.py` if you'd like
* The intended conventions here are `UPPER_SNAKE_CASE` for ambient state and `python_naming_conventions`/`MSBuildNamingConventions` for variables passed to the respective system within the scope of the workflow.
* `create-portable-zip.py` didn't follow this convention, whoops. (Will fix with [#1881](https://github.com/bonsai-rx/bonsai/issues/1881))

Long version:

---------------------------------

> Is `os.getenv` case-insensitive?

It is case sensitive on Linux and case-insensitive on Windows. This is the correct case though, this is a weird clash of naming conventions where I used `UPPER_CASE_SNAKE` for the repository variable (which is a common convention and helps make it clear that it's ambient state) but I use the same thing as the Python variable name for passing variables to Python scripts (which is a personal convention as noted below.)

(So I think what my brain is saying is that `ENABLE_PACKAGE_COMPARISON` is ambient state associated with the repository until it is passed to the script after which it no longer is.)

> I can see that in the build workflow there is a lower snake case field `enable_package_comparison` definition in the env section, which probably is what is used here?

Correct, my convention* is that environment variables intended purely for scripts use the convention where they'll be primarily used. So variables intended for Python scripts are `lower_snake_case`, variables for MSBuild are `PascalCase` (EG: `IsReferenceDummyBuild`), etc.

I do this so they look natural where they're used but also it gives a hint that it's something specific to this repo and not ambient environment stuff. It also helps keep things aligned in cases where we don't define the variable. (`ContinuousIntegrationBuild` is an example of this, that's a .NET SDK thing that I did not name.)

Actions workflows are the glue between a lot of different systems so sometimes you get dogs and cats living together.

(*Your comment made me realize that `create-portable-zip.py` doesn't follow this convention though 😅 I might fix that when I address https://github.com/bonsai-rx/bonsai/issues/1881)

> However, I was still then left wondering about `GITHUB_EVENT_NAME` which is in upper snake case in the python script, but lower snake case in the env section and I couldn't find it defined elsewhere.

This one is ambient state from GitHub Actions, it's [documented here](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables).

Using it here was probably just a mood thing. I think originally it was because the script didn't take any environment variables before and I didn't want to add one so I just used the default available thing.

That does make it inconsistent with `configure-build.py` which has it passed manually. Not sure if that's because `configure-build.py` predates GitHub adding `GITHUB_EVENT_NAME` ([it's quite old](https://github.com/MochiLibraries/Biohazrd/blob/4dbaabfb8195853a931a21d9b6468baed9020bf0/.github/workflows/configure-build.py)) or if I wanted it to be explicitly passed so it's clear it influences the outcome of the script. (I suspect the former might be the original reason but the latter would be my reason for keeping it.)

I could change `create-build-matrix.py` to do the same as well for the latter reason/consistency if you'd like.

glopesdev: It's fine, no need to change and I agree with all the conventions, thanks for clarifying!

glopesdev (approved):

PR #1890, opened 2024-07-08
Preload all extension types from include workflows

### LINKED ISSUE #824: Editor crashes on file open when failing to load included workflow
To reproduce: open a workflow file containing missing included workflows, or invalid paths to embedded assemblies which cannot load (i.e. an included workflow pointing to a missing package).

This is a regression caused by fixing #800 with eager loading without accounting for potential exceptions raised during load. If this is the case, it might be okay to simply catch any exceptions to ensure `ArgumentRange` never throws, and wait for the error to be thrown during proper workflow building.

### LINKED ISSUE #1680: Performance issues when loading include workflows
A growing issue as we use more and more embedded resources is a performance penalty when accessing the first embedded resource in a newly loaded assembly. This penalty seems to be paid only once when the resource is first loaded, but is paid for every single assembly with embedded resources.

The first step here would be to investigate the performance penalty further with benchmarks on both .NET framework and .NET core, and with different types of resources and different numbers of assemblies and resources loaded (e.g. does an assembly with 1 embedded resource pay a smaller cost than an assembly with 10 embedded resources?).

glopesdev:
Include workflows are loaded lazily and on-demand during the workflow build step, which means extension type discovery is performed in step-fashion for a newly loaded workflow. This gradual type discovery has dramatic implications in performance as documented in #1680, and this was having a visible impact on editor loading times, especially for complex applications with large numbers of include workflows.

This PR aims to improve the performance by preloading all extension types from included workflows ahead of time. The serialization of workflow properties was also optimized to minimize the number of required serializer instances.

The preloading optimization is disabled during builds to avoid supra-linear costs when recursively loading included workflows with deeply nested self-repeating motifs where the same workflow might be referenced multiple times.

Fixes #1680 
Fixes #824

### DIFF
--- a/Bonsai.Core/Expressions/IncludeWorkflowBuilder.cs
+++ b/Bonsai.Core/Expressions/IncludeWorkflowBuilder.cs
@@ -344,12 +344,11 @@ namespace Bonsai.Expressions
             }
             else
             {
-                if (!File.Exists(path))
+                try { return File.OpenRead(path); }
+                catch (FileNotFoundException ex)
                 {
-                    throw new InvalidOperationException("The specified workflow could not be found.");
+                    throw new InvalidOperationException("The specified workflow could not be found.", ex);
                 }
-
-                return File.OpenRead(path);
             }
         }
 
--- a/Bonsai.Core/Expressions/IncludeWorkflowBuilder.cs
+++ b/Bonsai.Core/Expressions/IncludeWorkflowBuilder.cs
@@ -25,7 +25,7 @@ namespace Bonsai.Expressions
     {
         const char AssemblySeparator = ':';
         internal const string BuildUriPrefix = "::build:";
-        const string DefaultSerializerPropertyName = "Property";
+        const string PlaceholderSerializerPropertyName = "__Property__";
         static readonly XElement[] EmptyProperties = new XElement[0];
         static readonly XmlSerializerNamespaces DefaultSerializerNamespaces = GetXmlSerializerNamespaces();
 
@@ -243,7 +243,7 @@ namespace Bonsai.Expressions
             }
 
             var previousName = element.Name;
-            element.Name = XName.Get(DefaultSerializerPropertyName, element.Name.NamespaceName);
+            element.Name = XName.Get(PlaceholderSerializerPropertyName, element.Name.NamespaceName);
             try
             {
                 var serializer = PropertySerializer.GetXmlSerializer(property.PropertyType);
@@ -431,6 +431,9 @@ namespace Bonsai.Expressions
             return workflow.BuildNested(arguments, includeContext);
         }
 
+        // We serialize all properties using the same placeholder root name to allow us to reuse
+        // a single serializer for each different property type. This means we need to rename
+        // the actual XElement name before or after serialization and deserialization.
         static class PropertySerializer
         {
             static readonly Dictionary<Type, XmlSerializer> serializerCache = new();
@@ -443,7 +446,7 @@ namespace Bonsai.Expressions
                 {
                     if (!serializerCache.TryGetValue(type, out serializer))
                     {
-                        var xmlRoot = new XmlRootAttribute(DefaultSerializerPropertyName) { Namespace = Constants.XmlNamespace };
+                        var xmlRoot = new XmlRootAttribute(PlaceholderSerializerPropertyName) { Namespace = Constants.XmlNamespace };
                         serializer = new XmlSerializer(type, xmlRoot);
                         serializerCache.Add(type, serializer);
                     }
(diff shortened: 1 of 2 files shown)


Bonsai.Core/WorkflowBuilder.cs
```diff
@@ -888,6 +901,29 @@ static void WriteXmlAttributes(XmlReader reader, XmlWriter writer, bool lookupTy
...
                                     }
                                 }
+                                else if (includeWorkflow &&
+                                         reader.GetAttribute(nameof(IncludeWorkflowBuilder.Path)) is string path &&
+                                         !reader.BaseURI.StartsWith(IncludeWorkflowBuilder.BuildUriPrefix) &&
+                                         visitedWorkflows.Add(path))
+                                {
+                                    // we don't want to fail in most cases while reading nested metadata, as this
+                                    // is an optional performance optimization and we would lose the visual context
+                                    // as to where exactly in the workflow the failure is happening
```
PathogenDavid, line 911: Is this an actual concern or just being overly defensive since this is a pretty low-level change to a fundamental system?

Bonsai.Core/Expressions/IncludeWorkflowBuilder.cs
```diff
@@ -24,6 +24,8 @@ namespace Bonsai.Expressions
     public sealed class IncludeWorkflowBuilder : VariableArgumentExpressionBuilder, IGroupWorkflowBuilder, INamedElement, IRequireBuildContext
     {
         const char AssemblySeparator = ':';
+        internal const string BuildUriPrefix = "::build:";
+        const string DefaultSerializerPropertyName = "Property";
```
PathogenDavid, line 28: I've gathered that this is part of what allows using the same `XmlSerializer` for the same types in multiple unrelated contexts, but it took a while to get there.

Would `Placeholder` maybe be better than `Default` here? A clarifying comment that this is only used ephemerally to allow sharing `XmlSerializer` instances might be helpful too.

PathogenDavid, line 28: Something less generic than `Property` might be valuable here in case this somehow finds its way into an exception message or something. (Even something like `__Property__` would make it searchable.)

PathogenDavid (approved):
I'm limited in how much I can see far-reaching concerns here, but the changes look good to me. Just a few requests for clarity.

Bonsai.Core/WorkflowBuilder.cs
glopesdev: It's mostly a user-experience thing. It is definitely possible for workflows to have references to missing files or links to files which are actually corrupt, etc. The main issue is that if we let the loader of the main workflow crash when trying to inspect those files, we will have no way of telling visually in which part of the internal workflow the reference is located (e.g. it could be located in a particular node two-levels deep).

This is because the eager lookahead is done without actually loading up and deserializing the XML, so we don't know about internal structure at this point. If there is an exception like this, it will also crash later when the included workflow is actually built and at that time the same exception will be thrown but at the right place in the workflow.

glopesdev:
@PathogenDavid I ended up doing one final refactor to defer workflow assignment as you suggested in #824. Turns out the fix required significant groundwork to avoid depending on `TypeDescriptor` calls using `this`.

The refactoring in itself is definitely an improvement. I was a bit worried something subtle might break, but thankfully the most complicated infrastructure was already in place and we have some test coverage of some of the most complex edge cases which got triggered and fixed. I have also tested usage with some fairly complex workflows and everything seems to work fine.

Bonsai.Core/Expressions/IncludeWorkflowBuilder.cs
```diff
@@ -381,29 +387,26 @@ void EnsureWorkflow(IBuildContext buildContext)
...
                     }
 
                     if (properties != null)
                     {
-                        SetXmlProperties(properties);
+                        SetXmlProperties(builder.Workflow, properties);
                     }
+
+                    var parameterCount = builder.Workflow.GetNestedParameters().Count();
+                    workflow = inspectWorkflow ? builder.Workflow.ToInspectableGraph() : builder.Workflow;
```
PathogenDavid, line 406: Nothing following this is likely to throw, but this should still really be the very last thing assigned since it's the sentry value.

PathogenDavid, line 406: Nothing following this is likely to throw, but this should still really be the very last thing assigned since it's the sentry value.

PathogenDavid (changes requested):

Bonsai.Core/Expressions/IncludeWorkflowBuilder.cs
PathogenDavid: The idea is that we shouldn't have to think about it this much in the first place, but fair enough with the argument range.

Ideally this should all be encapsulated better so that the object can't even exist in an invalid state at all, but that'd be a pretty major refactoring that's way out of scope for this PR.

PathogenDavid (approved):

PR #1893, opened 2024-07-08
Use forward slashes for package configuration

### LINKED ISSUE #1864: Environment config files are not compatible across platforms
Currently the local environment `.config` file relies on OS-specific path separators to determine assembly locations and native library folders. Because of this, it is currently not possible to version `.config` files across platforms, e.g. Windows and Linux, since the direction of slashes is different.

A proposal was made in #1572 to resolve this by simply removing the explicit locations from the config file entirely. Unfortunately there are challenges to implementing this quickly, so it might be easier for now to simply normalize the paths to the current OS when reading, and normalize to a fixed convention when writing the config file, e.g. to linux-style for increased compatibility.

glopesdev:
The local environment `.config` file has relied so far on OS-specific path strings to determine assembly locations and native library folders. However, because of this it is currently not possible to version `.config` files across platforms, e.g. Windows and Linux, since the direction of slashes is different.

This PR normalizes config file manipulations to prefer Unix paths for all package install, update and removal operations. To allow for the greatest backwards compatibility as well as OS-specific path strings passed through the CLI interface, the behavior of the `PackageConfiguration` and `ConfigurationHelper` classes was not modified. Instead, normalization was implemented via the `PackageConfigurationUpdater`, which is responsible for actually manipulating the serializable representation of the `.config` file.

When initializing the updater object, all existing paths are normalized to Unix forward slashes, and any further modifications are then performed and persisted using forward slash path separators.

Fixes #1864

### DIFF
--- a/Bonsai.Configuration/PackageConfigurationUpdater.cs
+++ b/Bonsai.Configuration/PackageConfigurationUpdater.cs
@@ -46,7 +46,7 @@ namespace Bonsai.Configuration
         public PackageConfigurationUpdater(NuGetFramework projectFramework, PackageConfiguration configuration, IPackageManager manager, string bootstrapperPath = null, PackageIdentity bootstrapperName = null)
         {
             packageManager = manager ?? throw new ArgumentNullException(nameof(manager));
-            packageConfiguration = NormalizePathSeparators(configuration ?? throw new ArgumentNullException(nameof(configuration)));
+            packageConfiguration = configuration ?? throw new ArgumentNullException(nameof(configuration));
             bootstrapperFramework = projectFramework ?? throw new ArgumentNullException(nameof(projectFramework));
             bootstrapperExePath = bootstrapperPath ?? string.Empty;
             bootstrapperDirectory = Path.GetDirectoryName(bootstrapperExePath);
@@ -58,6 +58,7 @@ namespace Bonsai.Configuration
             var galleryPath = Path.Combine(bootstrapperDirectory, GalleryDirectory);
             var galleryPackageSource = new PackageSource(galleryPath);
             galleryRepository = new SourceRepository(galleryPackageSource, Repository.Provider.GetCoreV3());
+            NormalizePathSeparators(packageConfiguration);
         }
 
         string GetRelativePath(string path)
@@ -74,29 +75,21 @@ namespace Bonsai.Configuration
             return PathUtility.GetPathWithForwardSlashes(Path.Combine(path1, path2));
         }
 
-        static PackageConfiguration NormalizePathSeparators(PackageConfiguration configuration)
+        static void NormalizePathSeparators(PackageConfiguration configuration)
         {
-            var normalized = new PackageConfiguration();
-            normalized.ConfigurationFile = configuration.ConfigurationFile;
-            normalized.Packages.AddRange(configuration.Packages);
-            normalized.AssemblyReferences.AddRange(configuration.AssemblyReferences);
-
             foreach (var assemblyLocation in configuration.AssemblyLocations)
             {
-                normalized.AssemblyLocations.Add(
-                    assemblyLocation.AssemblyName,
-                    assemblyLocation.ProcessorArchitecture,
-                    PathUtility.GetPathWithForwardSlashes(assemblyLocation.Location));
+                assemblyLocation.Location = PathUtility.GetPathWithForwardSlashes(assemblyLocation.Location);
             }
 
-            foreach (var folder in configuration.LibraryFolders)
+            // cannot normalize in place since path is collection key
+            var libraryFolders = configuration.LibraryFolders.ToArray();
+            configuration.LibraryFolders.Clear();
+            foreach (var folder in libraryFolders)
             {
-                normalized.LibraryFolders.Add(
-                    PathUtility.GetPathWithForwardSlashes(folder.Path),
-                    folder.Platform);
+                folder.Path = PathUtility.GetPathWithForwardSlashes(folder.Path);
+                configuration.LibraryFolders.Add(folder);
             }
-
-            return normalized;
         }
 
         static bool IsTaggedPackage(PackageReaderBase package)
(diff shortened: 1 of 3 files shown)


Bonsai.Configuration/PackageConfigurationUpdater.cs
```diff
@@ -69,6 +69,36 @@ string GetRelativePath(string path)
             return PathUtility.GetPathWithDirectorySeparator(relativeUri.ToString());
         }
 
+        static string CombinePath(string path1, string path2)
+        {
+            return PathUtility.GetPathWithForwardSlashes(Path.Combine(path1, path2));
+        }
+
+        static PackageConfiguration NormalizePathSeparators(PackageConfiguration configuration)
```
PathogenDavid, line 77: This method feels like it belongs on `PackageConfiguration` itself. (Maybe as `WithNormalizedPathSeparators`.)

As it stands right now, there's a big risk of someone overlooking this method and not knowing that it needs to be updated if they were adding a field to `PackageConfiguration`.

Bonsai.Configuration/PackageConfigurationUpdater.cs
```diff
@@ -69,6 +69,36 @@ string GetRelativePath(string path)
...
+        {
+            var normalized = new PackageConfiguration();
+            normalized.ConfigurationFile = configuration.ConfigurationFile;
```
PathogenDavid, line 80: Even though it doesn't appear to actually matter, this should be normalized too because otherwise it looks like a bug at first glance.

Bonsai.Configuration/PackageConfigurationUpdater.cs
```diff
@@ -69,6 +69,36 @@ string GetRelativePath(string path)
...
+                    PathUtility.GetPathWithForwardSlashes(assemblyLocation.Location));
+            }
+
+            foreach (var folder in configuration.LibraryFolders)
+            {
+                normalized.LibraryFolders.Add(
+                    PathUtility.GetPathWithForwardSlashes(folder.Path),
+                    folder.Platform);
+            }
+
```
PathogenDavid, line 98: It doesn't appear to actually matter, but `ExtensionFolders` should get copied+normalized too. (It should at least be copied over.)

Bonsai.Configuration/PackageConfigurationUpdater.cs
PathogenDavid: This would also make it more reasonable to unit test this behavior in isolation if you have time for a quick test.

PathogenDavid (changes requested):

PathogenDavid:
Also glad to see the eradication of `System.Tuple` usage :)

Bonsai.Configuration/PackageConfigurationUpdater.cs
glopesdev: Point taken about unit testing, but the rationale behind it is that we actually don't want to normalize all paths in the configuration (see below), and it also might mislead developers that this is somehow a desirable behavior to do, whereas actually there is only a limited set of cases for it.

Also in my case it helps to know that all responsibility for path normalization is centered in a single class. More details in the below comment, but happy to revisit this.

Bonsai.Configuration/PackageConfigurationUpdater.cs
glopesdev: The `ConfigurationFile` property is not normalized since 1) it doesn't get persisted on file (`[XmlIgnore]`); and 2) it may refer to an absolute path, rather than a relative path. I am trying to avoid here the reverse Unix-problem where we might be unable to refer to specific paths on Windows because we normalize all backslashes (e.g. for a configuration file or bonsai environment loaded from a shared network drive folder).

My overall reasoning here is that it is always safe to apply path normalization if the path is a relative path, but not if it is an absolute path.

Bonsai.Configuration/PackageConfigurationUpdater.cs
glopesdev: Same reasoning as above for `ConfigurationFile`. In addition the reason for not copying `ExtensionFolders` is that they are explicitly out of scope of the `PackageConfigurationUpdater` and should never be touched by it or risk saving anything to file.

glopesdev: I was giving it another look and it might be the case that we can't actually copy the `PackageConfiguration` object at all, but rather try to modify it in place. The reason is that the original object gets passed over in sequence to different asynchronous package recovery updaters and they were supposed to fill it with restored packages, whereas now this will never happen.

Bonsai.Configuration/PackageConfigurationUpdater.cs
PathogenDavid: Looking at it again I agree, especially since normalizing the `ConfigurationFile` is definitely actually a bad thing.

Bonsai.Configuration/PackageConfigurationUpdater.cs
PathogenDavid: Agreed, doing so would break UNC paths.

Bonsai.Configuration/PackageConfigurationUpdater.cs
PathogenDavid: Just to clarify: You're saying you were originally trying to avoid modifying the existing `PackageConfiguration` but now you're OK with it? (I agree it seems like it should be fine, and the revised implementation indirectly addresses my concerns with this getting out of sync with `PackageConfiguration`.)

PathogenDavid (approved):
Looks good to me!

Bonsai.Configuration/PackageConfigurationUpdater.cs
glopesdev: Yup, that's right. One of those cases where it seemed like potentially a good idea but actually isn't 😅 (in the current code base at least). We can revisit all this later but out of scope of this patch.

PR #1894, opened 2024-07-09
Hide embedded DLLs from the solution explorer

PathogenDavid:
Another minor echo of the release automation changes.

This hides the DLLs embedded in the bootstrapper for `System.Resources.Extensions` and its dependencies from the solution explorer, which are unintentionally left visible before this change:

Screenshot of Visual Studio solution explorer with DLL files listed

These don't live in the project folder, and without the context of reading the csproj they're just confusing noise.

(That being said if you like being reminded that they're there we can just leave them. Could also hide them in a folder as an in-between option.)

Relates to https://github.com/bonsai-rx/bonsai/pull/1873


glopesdev (approved):
I agree in this case it makes more sense to hide them since there are a bunch of other references which are being repacked as well and do not show up.

Also hopefully we are migrating away from this in the near future so we won't need the reminder.

Issue #1901, opened 2024-07-11
Packaged release bootstrapper fails on mono


glopesdev:
Running the latest packaged release of the bootstrapper on Linux mono fails with the below stack trace. Compiling the project from source with `dotnet build` is successful however, so the suspicion is this might be possibly related to ILRepack and embedding of `System.Resources.Extensions`.

The current workaround is to build from source.

```
Unhandled Exception:
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.ArgumentException: This .resources file should not be read with this reader. The resource reader type is "System.Resources.Extensions.DeserializingResourceReader".
Parameter name: reader
  at System.Resources.Extensions.RuntimeResourceSet..ctor (System.Resources.IResourceReader reader) [0x0002f] in <efe718142b824e97b598670ead64f4bf>:0 
  at (wrapper managed-to-native) System.Reflection.RuntimeConstructorInfo.InternalInvoke(System.Reflection.RuntimeConstructorInfo,object,object[],System.Exception&)
  at System.Reflection.RuntimeConstructorInfo.InternalInvoke (System.Object obj, System.Object[] parameters, System.Boolean wrapExceptions) [0x00005] in <d636f104d58046fd9b195699bcb1a744>:0 
   --- End of inner exception stack trace ---
  at System.Reflection.RuntimeConstructorInfo.InternalInvoke (System.Object obj, System.Object[] parameters, System.Boolean wrapExceptions) [0x0001a] in <d636f104d58046fd9b195699bcb1a744>:0 
  at System.Reflection.RuntimeConstructorInfo.DoInvoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00086] in <d636f104d58046fd9b195699bcb1a744>:0 
  at System.Reflection.RuntimeConstructorInfo.Invoke (System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00000] in <d636f104d58046fd9b195699bcb1a744>:0 
[...]

PR #1902, opened 2024-07-12
Fix redundant assembly resolution resulting in duplicate assembly loading

PathogenDavid:
Also added copying the `NuGet.config` file to the repack output directory for ease of testing.

Fixes https://github.com/bonsai-rx/bonsai/issues/1901

### DIFF
--- a/Bonsai/SystemResourcesExtensionsSupport.cs
+++ b/Bonsai/SystemResourcesExtensionsSupport.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Diagnostics;
 using System.Reflection;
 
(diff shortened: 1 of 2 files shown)


Bonsai/SystemResourcesExtensionsSupport.cs
```diff
@@ -13,22 +14,30 @@ namespace Bonsai
     /// </remarks>
     internal static class SystemResourcesExtensionsSupport
     {
+        // Assembly.Load(byte[]) always results in a new assembly so we need to manually de-duplicate them
+        private static ConcurrentDictionary<string, Assembly> resolveCache = new();
```
glopesdev, line 18: Do we really need the `ConcurrentDictionary` here? My understanding was that the `AssemblyResolve` event is never called concurrently, which is why I used a normal `Dictionary` in `ConfigurationHelper`.

I was trying to find a reference for this on the [`AssemblyResolve`](https://learn.microsoft.com/en-us/dotnet/api/system.appdomain.assemblyresolve?view=net-8.0) docs and in the [Resolve assembly loads guidance](https://learn.microsoft.com/en-us/dotnet/standard/assembly/resolve-loads) but could not find anything conclusive unfortunately, although I would indeed expect that if this is an issue someone would have flagged it. There is a [SO question](https://stackoverflow.com/questions/9180638/working-with-appdomain-assemblyresolve-event#9180843) that also uses a simple `IDictionary` and again no comments decrying its use, but surprised that I cannot find something definitive on this.

If it turns out to be the case we do need it, I would say we should probably change this in `ConfigurationHelper` as well, probably even as part of this same PR.

glopesdev, line 18: Separately, I was wondering whether we should move `resolveCache` inside the body of the `Initialize` method? This would prevent initializing an unused concurrent dictionary in .NET Core since it would be guarded by the `ConditionalAttribute`.

glopesdev: Decided to adopt caching for all assembly resolution in #1904.

PathogenDavid: (For the sake of posterity: I messaged Gonçalo on Teams to let him know `ConfigurationHelper` had this problem before I saw this and we talked about it there.)

> I am surprised by their choice of returning `null` in case the assembly is already in the process of being resolved, since it seems this might result in assembly resolution exceptions potentially being thrown from weird places at weird times.

In this case I think that behavior is very specific to the recursion issue they're working around. They actually intentionally want the inner resolves to fail.

PathogenDavid: Good idea. That mildly annoyed me too but not long enough for me to think of that solution 😅

glopesdev (approved):

PR #1910, opened 2024-07-14
Fix exception constructor argument order

glopesdev:
For some reason the order of arguments into the [`ArgumentOutOfRangeException`](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception.-ctor?view=net-8.0#system-argumentoutofrangeexception-ctor(system-string-system-string)) constructor was swapped in all instances of the call in the `DirectedGraph` class, most likely because it is the reverse order from the more common [`ArgumentException`](https://learn.microsoft.com/en-us/dotnet/api/system.argumentexception.-ctor?view=net-8.0#system-argumentexception-ctor(system-string-system-string)).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -34,3 +34,6 @@ dotnet_sort_system_directives_first = true
 # Predefined for members, etc does not create a message because the explicitly sized types are conveient in interop scenarios where the bit size matters.
 dotnet_style_predefined_type_for_locals_parameters_members = true:none
 dotnet_style_predefined_type_for_member_access = true:suggestion
+
+# Instantiate argument exceptions correctly
+dotnet_diagnostic.CA2208.severity = warning
--- a/Bonsai.Core/Dag/DirectedGraph.cs
+++ b/Bonsai.Core/Dag/DirectedGraph.cs
@@ -191,7 +191,7 @@ namespace Bonsai.Dag
             ThrowIfEdgeVerticesNullOrNotInGraph(from, to, nameof(to));
             if (edgeIndex < 0 || edgeIndex > from.Successors.Count)
             {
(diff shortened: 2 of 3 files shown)


PathogenDavid (approved):
Good as-is, but it might be a good idea to swap most of these over to the [`ArgumentOutOfRngeException(string paramName)` overload](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception.-ctor?view=net-8.0#system-argumentoutofrangeexception-ctor(system-string)) instead since their messages don't really add anything over the default one. (That overload is why the order is swapped from the base `ArgumentException`.)

-----------------------

It also might be worth enabling CA2208 to avoid this mistake in the future:

```editorconfig
# Add to .editorconfig
dotnet_diagnostic.CA2208.severity = warning
```

Although it's worth noting that'll complain about [inappropriate usage of `ArgumentOutOfRangeException` in `Slice.cs`](https://github.com/bonsai-rx/bonsai/blob/3464e656adf0c6354be1c98e5877d40e0e94adfb/Bonsai.Core/Reactive/Slice.cs#L46-L62). (Because they're locals rather than parameters, but I think it makes reasonable sense in a Bonsai context--could just suppress.)

glopesdev:
> Although it's worth noting that'll complain about [inappropriate usage of ArgumentOutOfRangeException in Slice.cs](https://github.com/bonsai-rx/bonsai/blob/3464e656adf0c6354be1c98e5877d40e0e94adfb/Bonsai.Core/Reactive/Slice.cs#L46-L62). (Because they're locals rather than parameters, but I think it makes reasonable sense in a Bonsai context--could just suppress.)

Actually that was indeed another inappropriate use, so I've turned diagnostic severity to warning. We have been using `InvalidOperationException` rather than `ArgumentException` for these, but `Slice` is old enough that it retained some legacy.

PR #1913, opened 2024-07-15
Drop version build metadata for official releases

### LINKED ISSUE #1912: Avoid displaying commit hash in AboutBox and workflow file version
This probably makes sense for dev builds, but I think it might be confusing especially for people versioning workflows to suddenly have the git hash come up for full releases.

I think this currently is picked up only in two places in `WorkflowBuilder` and `AboutBox` respectively:

https://github.com/bonsai-rx/bonsai/blob/3464e656adf0c6354be1c98e5877d40e0e94adfb/Bonsai.Core/WorkflowBuilder.cs#L204

https://github.com/bonsai-rx/bonsai/blob/3464e656adf0c6354be1c98e5877d40e0e94adfb/Bonsai.Editor/AboutBox.cs#L63

In both cases we can condition on `BUILD_KIND_OFFICIAL_RELEASE` to decide how to assemble the version string.

glopesdev:
To avoid cluttering official release version strings stored in workflow files and displayed in the About box, this PR drops source link build metadata from product version strings.

Fixes #1912

### DIFF
--- a/Bonsai.Core/WorkflowBuilder.cs
+++ b/Bonsai.Core/WorkflowBuilder.cs
@@ -200,8 +200,14 @@ namespace Bonsai
             writer = new XmlExtensionWriter(writer, genericTypes);
 
             var assembly = Assembly.GetExecutingAssembly();
-            var versionInfo = FileVersionInfo.GetVersionInfo(assembly.Location);
-            writer.WriteAttributeString(VersionAttributeName, versionInfo.ProductVersion);
+            var version = FileVersionInfo.GetVersionInfo(assembly.Location).ProductVersion;
+#if BUILD_KIND_OFFICIAL_RELEASE
+            // Drop build metadata for official releases
+            var plus = version.IndexOf('+');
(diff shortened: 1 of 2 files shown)


Bonsai.Core/WorkflowBuilder.cs
```diff
@@ -220,7 +226,7 @@ void IXmlSerializable.WriteXml(XmlWriter writer)
             serializer.Serialize(writer, workflow.ToDescriptor(), serializerNamespaces);
         }
 
-        #endregion
+#endregion
```
PathogenDavid, line 229: Nit: Seemingly unintended indentation change. (Region blocks don't usually get left-adjusted and aren't elsewhere in this file.)

PathogenDavid (approved):

Bonsai.Core/WorkflowBuilder.cs
glopesdev: Good catch!

PR #1924, opened 2024-07-16
Update ScintillaNET to support modern .NET

### LINKED ISSUE #1706: Consider moving annotation editor to editor package
Currently the enhanced annotation editor is implemented in the `Bonsai.Design` package. This is not ideal, as it brings in specific dependencies, i.e. Scintilla.NET which end up being unnecessarily propagated to downstream consumers of the design infrastructure.

A better solution would be to have this editor implemented as part of the `Bonsai.Editor` package, and change the `AnnotationBuilder` component editor to reference this new editor type; or simply remove the editor attribute altogether and have the component editor be injected via the workflow editor form itself. This would work since the `AnnotationBuilder` is a known core type which the editor assembly can know about.

glopesdev:
In preparation for modernizing the editor to run with the .NET Core stack there are a number of core dependencies which need to be upgraded or modified to run beyond .NET Framework. [jacobslusser/ScintillaNET](https://github.com/jacobslusser/ScintillaNET) was until now one of the big concerns, since the project has been stale for almost 6+ years.

Fortunately, [fernandreu/ScintillaNET](https://github.com/fernandreu/ScintillaNET) now provides a .NET 6.0 compatible fork with a published NuGet package and the source code updates seem to be in great shape. Even though it still only targets windows, the component has even been updated with a new `arm64` runtime, which is a welcome addition for possibly targeting Windows on ARM.

Since this fork only targets .NET 4.7.1 we took the chance to update all projects to target .NET 4.7.2 at a minimum, which is now our recommended target for new packages anyway.

Closes #1706

### DIFF
--- a/Bonsai.Audio/Bonsai.Audio.csproj
+++ b/Bonsai.Audio/Bonsai.Audio.csproj
@@ -3,7 +3,7 @@
     <Title>Bonsai - Audio Library</Title>
     <Description>Bonsai Audio Library containing modules for sound capture and playback.</Description>
     <PackageTags>Bonsai Rx Audio</PackageTags>
-    <TargetFramework>net462</TargetFramework>
+    <TargetFramework>net472</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="openal.redist" Version="2.0.7" />
--- a/Bonsai.Core.Tests/Bonsai.Core.Tests.csproj
+++ b/Bonsai.Core.Tests/Bonsai.Core.Tests.csproj
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <IsPackable>false</IsPackable>
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
-    <TargetFramework>net462</TargetFramework>
+    <TargetFramework>net472</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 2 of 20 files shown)


PathogenDavid (approved):
Does this PR actually fix https://github.com/bonsai-rx/bonsai/issues/1706? That issue seems to be primarily about moving this dependency out of `Bonsai.Design`, which is not what's being done here. (The comment is about that, but the comment also seems mostly unrelated to the main issue?)

Other than that the PR looks good.

glopesdev:
Good point, I have amended the wording to "closes".

PathogenDavid:
@glopesdev To clarify: I don't see how this PR addresses https://github.com/bonsai-rx/bonsai/issues/1706 at all. Fixing closing or otherwise. It addresses the comment, but it's also not clear how that comment really relates to the actual issue 😅

Migrating to the new package doesn't really solve the issue of imposing this package on downstream consumers of `Bonsai.Design`. Was https://github.com/bonsai-rx/bonsai/issues/1706 intended primarily as paving a path forward for modern .NET support rather than fixing the unnecessary transitive dependency?

glopesdev:
@PathogenDavid agreed, I was distracted by another shared dependency `RichTextEditor` (long story) but it ends up not relating to this upstream dependency at all, the issue still stands, will reopen.

PR #1931, opened 2024-07-20
Allow propagation of Defer nested sequence visualizer

### LINKED ISSUE #1896: Consider allowing Defer to propagate internal sequence visualizer
I just realized that there may actually be a 3rd case

## 3. Propagation of internal scope sequence + visualizer

For instance, consider the following workflow:

image

A potential pattern might be to nest these operators within a `Defer` and want the visualizer output to be `Button` while the output to be the `Int`. I wonder if a more "Bonsonic" way to do this is to allow `VisualizerMapping` to target `WorkflowOutputs`. For instance:

image

_Originally posted by @bruno-f-cruz in https://github.com/bonsai-rx/bonsai/issues/1860#issuecomment-2192593731_

glopesdev:
Previous enhancements to visualizer propagation rules introduced in #917, #1751 and #1889 allow flexible reuse, modification and encapsulation of visualizers inside group workflows and across subject multicasting.

However, some patterns remain problematic to achieve without introducing additional subject declarations. Specifically, given an observable sequence where data is generated by interaction with a UI element, it is not currently easy to post-process the values in the sequence without either losing the assigned visualizer or introducing a temporary subject.

This PR allows propagation of the visualizer element associated with the nested output in `Defer` operators, which enables encapsulation of these more flexible constructions with subjects.

For example, by combining `Defer` with `Visualizer`, and leveraging the ability to chain `VisualizerMapping` nodes into the lifted nested visualizer, it becomes possible to solve the case problem discussed in #1896:

#### Workflow
image

#### Button (Defer)
image

#### Visualizer
image

<details>
  <summary>Copy workflow</summary>

```xml
<?xml version="1.0" encoding="utf-8"?>
<WorkflowBuilder Version="2.8.6-dev0+9f6565675c7a956a370724f1022102862a05eb1d"
                 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                 xmlns:rx="clr-namespace:Bonsai.Reactive;assembly=Bonsai.Core"
                 xmlns:gui="clr-namespace:Bonsai.Gui;assembly=Bonsai.Gui"
                 xmlns="https://bonsai-rx.org/2018/workflow">
  <Workflow>
    <Nodes>
      <Expression xsi:type="rx:Defer">
        <Name>Button</Name>
        <Workflow>
          <Nodes>
            <Expression xsi:type="SubscribeSubject">
              <Name>ButtonPressed</Name>
            </Expression>
            <Expression xsi:type="Combinator">
              <Combinator xsi:type="IntProperty">
                <Value>0</Value>
              </Combinator>
            </Expression>
            <Expression xsi:type="gui:ButtonBuilder">
              <gui:Enabled>true</gui:Enabled>
              <gui:Visible>true</gui:Visible>
              <gui:Text>Click</gui:Text>
            </Expression>
            <Expression xsi:type="rx:PublishSubject">
              <Name>ButtonPressed</Name>
            </Expression>
            <Expression xsi:type="VisualizerMapping" />
            <Expression xsi:type="rx:Visualizer">
              <Workflow>
                <Nodes>
                  <Expression xsi:type="gui:TableLayoutPanelBuilder">
                    <gui:Enabled>true</gui:Enabled>
                    <gui:Visible>true</gui:Visible>
                    <gui:ColumnCount>1</gui:ColumnCount>
                    <gui:RowCount>1</gui:RowCount>
                    <gui:ColumnStyles />
                    <gui:RowStyles />
                    <gui:CellSpans />
                  </Expression>
                  <Expression xsi:type="WorkflowOutput" />
                </Nodes>
                <Edges>
                  <Edge From="0" To="1" Label="Source1" />
                </Edges>
              </Workflow>
            </Expression>
            <Expression xsi:type="WorkflowOutput" />
          </Nodes>
          <Edges>
            <Edge From="0" To="1" Label="Source1" />
            <Edge From="1" To="5" Label="Source1" />
            <Edge From="2" To="3" Label="Source1" />
            <Edge From="3" To="4" Label="Source1" />
            <Edge From="4" To="5" Label="Source2" />
            <Edge From="5" To="6" Label="Source1" />
          </Edges>
        </Workflow>
      </Expression>
      <Expression xsi:type="MemberSelector" />
    </Nodes>
    <Edges>
      <Edge From="0" To="1" Label="Source1" />
    </Edges>
  </Workflow>
</WorkflowBuilder>
```
</details>

Fixes #1896

### DIFF
--- a/Bonsai.Core.Tests/InspectBuilderTests.cs
+++ b/Bonsai.Core.Tests/InspectBuilderTests.cs
@@ -101,6 +101,27 @@ namespace Bonsai.Core.Tests
         }
 
         [TestMethod]
+        public void Build_DeferInspectBuilder_ReturnNestedVisualizerElement()
+        {
+            // related to https://github.com/bonsai-rx/bonsai/issues/1896
+            ExpressionBuilder target = null;
+            var workflow = new TestWorkflow()
+                .AppendUnit()
+                .AppendNested(
+                    input => input
+                        .AppendUnit()
+                        .Capture(out target)
+                        .AppendOutput(),
+                    workflow => new Reactive.Defer(workflow))
+                .ToInspectableGraph();
+            workflow.Build();
+
+            var output = workflow[workflow.Count - 1].Value;
+            var visualizerElement = ExpressionBuilder.GetVisualizerElement(output);
+            Assert.AreSame(target, visualizerElement.Builder);
+        }
+
+        [TestMethod]
         public void Build_PropertyMappedInspectBuilderToWorkflowOutput_ReturnVisualizerElement()
         {
             ExpressionBuilder target;
--- a/Bonsai.Core.Tests/InspectBuilderTests.cs
+++ b/Bonsai.Core.Tests/InspectBuilderTests.cs
@@ -122,6 +122,33 @@ namespace Bonsai.Core.Tests
         }
 
         [TestMethod]
+        public void Build_DeferWithNestedVisualizerInspectBuilder_ReturnNestedVisualizerElement()
+        {
+            // related to https://github.com/bonsai-rx/bonsai/issues/1896
+            ExpressionBuilder target = null;
+            var workflow = new TestWorkflow()
+                .AppendUnit()
+                .AppendNested(
+                    input => input
+                        .AppendUnit()
+                        .AppendSubject<Reactive.BehaviorSubject>(nameof(System.Reactive.Unit))
+                        .AppendNested(
+                            input => input
+                                .AppendUnit()
+                                .Capture(out target)
+                                .AppendOutput(),
+                            workflow => new Reactive.Visualizer(workflow))
+                        .AppendOutput(),
+                    workflow => new Reactive.Defer(workflow))
+                .ToInspectableGraph();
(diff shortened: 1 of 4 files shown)


glopesdev:
@bruno-f-cruz @PathogenDavid given the discussion today and the realization that it is possible to map other visualizers directly onto the lifted output of the nested `Visualizer` operator, I don't think we need to do anything more for this release to resolve this issue.

PathogenDavid:
Do you want to mark this PR as draft then?

glopesdev:
@PathogenDavid actually, it's the contrary, we do want to go forward with this PR 👍 what was happening earlier is that I was wondering whether to add something to it (or make an alternative one) to cover the actual problem raised in the issue, but now we figured out that with just what is in this PR we have everything we need.

The bit with `Defer` is still important since it allows us to close off the scope around the inner subject, so you can recycle the whole thing inside a single `IncludeWorkflow` and make your own `Button` that outputs `int` instead of a `string` for example.

bruno-f-cruz (approved):

PathogenDavid (approved):

PR #1937, opened 2024-07-23
Replace uses of WebRequest with HttpClient

### LINKED ISSUE #1936: Consider replacing uses of WebRequest with HttpClient
The `WebRequest` APIs have been marked as obsolete from .NET 6.0 onwards. They are currently used in only two places in the code base (in [`PackageViewController`](https://github.com/bonsai-rx/bonsai/blob/69f4267012f017af8b0db3a3c9d427d1c45b4796/Bonsai.NuGet.Design/PackageViewController.cs#L251) and [`DocumentationHelper`](https://github.com/bonsai-rx/bonsai/blob/69f4267012f017af8b0db3a3c9d427d1c45b4796/Bonsai.Editor/DocumentationHelper.cs#L57)). Some loose guidelines for replacing these deprecated APIs may be found in the references below:

- https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-8.0
- https://stackoverflow.com/questions/70185058/how-to-replace-obsolete-webclient-with-httpclient-in-net-6

glopesdev:
The `WebRequest` API has sadly been deprecated by the more performant `HttpClient` in modern .NET. As this API is also available in .NET Framework, this PR aligns the two code bases by replacing all instances of `WebRequest` with an alternative implementation using `HttpClient`.

There were two main uses of `WebRequest` in the current source code:
- Retrieving package icons in the package manager dialog
- Retrieve XRefMap YAML content for package docs

Each of these is addressed separately, as the reimplementation is quite specific for each case. An opportunity was taken to refactor package icon retrieval away from UI code to improve separation of concerns.

Fixes #1936

### DIFF
--- a/Bonsai.NuGet.Design/IconReader.cs
+++ b/Bonsai.NuGet.Design/IconReader.cs
@@ -13,9 +13,8 @@ namespace Bonsai.NuGet.Design
 {
     class IconReader
     {
-        static readonly HttpClient httpClient = new();
-        static readonly Uri PackageDefaultIconUrl = new Uri("https://www.nuget.org/Content/Images/packageDefaultIcon.png");
-        static readonly TimeSpan DefaultIconTimeout = TimeSpan.FromSeconds(10);
+        static readonly HttpClient httpClient = GetHttpClient();
+        static readonly Uri PackageDefaultIconUrl = new("https://www.nuget.org/Content/Images/packageDefaultIcon.png");
         static readonly Image DefaultIconImage = new Bitmap(32, 32, PixelFormat.Format32bppArgb);
         readonly ConcurrentDictionary<Uri, Lazy<Task<Image>>> iconCache = new();
         readonly Task<Image> defaultIcon;
@@ -25,6 +24,14 @@ namespace Bonsai.NuGet.Design
         {
             targetSize = size;
             defaultIcon = GetAsync(PackageDefaultIconUrl);
+        }
+
+        static HttpClient GetHttpClient()
+        {
+            var client = new HttpClient();
+            client.DefaultRequestHeaders.CacheControl = new() { MustRevalidate = true };
+            client.Timeout = TimeSpan.FromSeconds(10);
+            return client;
         }
 
         public void ClearCache()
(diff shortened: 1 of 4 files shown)


Bonsai.NuGet.Design/IconReader.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            return await defaultIcon;
+        }
+
+        async Task<Image> GetWebRequestAsync(Uri requestUri, Task<Image> fallbackImage = null, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                using var requestCancellation = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+                requestCancellation.CancelAfter(DefaultIconTimeout);
```
PathogenDavid, line 75: Is there a reason this isn't using `HttpClient.Timeout` like `DocumentationHelper`?

Bonsai.NuGet.Design/IconReader.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using NuGet.Packaging;
+
+namespace Bonsai.NuGet.Design
+{
+    class IconReader
+    {
+        static readonly HttpClient httpClient = new();
```
PathogenDavid, line 16: It wasn't in the original implementation, but it seems like it'd be a good idea to enable the HTTP cache here too just like it is in `DocumentationHelper`.

(Compression *could* be enabled too, but it doesn't really matter in this context. NuGet.org won't bother trying to further compress the PNGs anyway, which makes sense.)

Bonsai.NuGet.Design/PackageViewController.cs
```diff
@@ -227,68 +217,9 @@ static Bitmap ResizeImage(Image image, Size newSize)
             return result;
```
PathogenDavid, line 217: `ResizeImage` was moved to `IconReader` and should be removed from `PackageViewController` as it is no longer referenced.

PathogenDavid (changes requested):
Just a few minor things

Bonsai.NuGet.Design/IconReader.cs
glopesdev: No reason at all, just that I only learned about it while refactoring `DocumentationHelper` 😅

Bonsai.NuGet.Design/PackageViewController.cs
glopesdev: Good catch!

PR #1941, opened 2024-07-30
Rename CombineTimestamp to CreateTimestamped

### LINKED ISSUE #1518: Consider renaming `CombineTimestamp` to `CreateTimestamped`
`CombineTimestamp` (Core) and `CreateTimestamped` (Bonsai.Harp) both perform the same conceptual function in that they take a primary value (`T`) and pair with a second input that represents some sort of temporal variable (`Timestamp`), to instantiate a `Timestamped<T>` type.

In order to create some UI predictability, and also considering that such a pattern might emerge in other upcoming packages, the suggestion is to uniformize the nomenclature and rename the core operator. 

For the sake of backward compatibility, the previous class should be marked as `[Obsolete]`.

glopesdev:
The `CombineTimestamp` operator was introduced in one of the earliest releases of the language as a way to create timestamped values dynamically from pairs of values and timestamps. As a consequence of this early development, the naming of this operator does not follow current naming conventions.

This PR refactors `CombineTimestamp` into the new `CreateTimestamped` operator, for consistency with other creation operators and with `CreateTimestamped` used in the Harp package. Core proxy types have been leveraged to allow for automatic conversion of old workflows and the original type has been marked with the `[Obsolete]` attribute.

Fixes #1518

### DIFF
--- a/Bonsai.Core/Reactive/CreateTimestamped.cs
+++ b/Bonsai.Core/Reactive/CreateTimestamped.cs
@@ -8,21 +8,21 @@ using System.ComponentModel;
 namespace Bonsai.Reactive
 {
     /// <summary>
-    /// Represents an operator that converts element-timestamp pairs of an observable
-    /// sequence into proper timestamped elements.
+    /// Represents an operator that converts element-timestamp pairs in an observable
+    /// sequence into <see cref="Timestamped{T}"/> values.
     /// </summary>
     [Combinator]
     [XmlType(Namespace = Constants.XmlNamespace)]
-    [Description("Converts a pair of element and timestamp into a proper timestamped type.")]
+    [Description("Converts a sequence of element-timestamp pairs into a sequence of timestamped values.")]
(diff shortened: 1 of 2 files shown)


Bonsai.Core/Reactive/CreateTimestamped.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Linq;
+using System.Reactive;
+using System.Reactive.Linq;
+using System.Xml.Serialization;
+using System.ComponentModel;
+
+namespace Bonsai.Reactive
+{
+    /// <summary>
+    /// Represents an operator that converts element-timestamp pairs of an observable
```
bruno-f-cruz, line 11: Consider referencing the rx Timestamped<T> struct here for href in the docs

bruno-f-cruz (approved):
Check the minor minor comment.

PathogenDavid (approved):

PR #1942, opened 2024-07-30
Allow building workflows as typed sequences

### LINKED ISSUE #1915: Consider adding BuildObservable overload that returns typed sequence
Several aspects of building and running workflows from code remain cumbersome. One of them is getting the result of compiled observable sequences, which currently has to be done by calling `Build` to get the workflow expression tree, and then assembling and calling a lambda to finally generate the typed sequence.

`BuildObservable` provides a wrapper to take care of this boilerplate code, but it currently assumes the type of the result sequence is always `Unit` by inserting a `UnitBuilder` at the end, which disregards any actual output value from the workflow.

Both for testing and other applications where the type of the workflow is either known or expected, we could provide a generic overload allowing for explicit indication of the type of the observable sequence returned by the workflow. This overload would test if the output sequence is compatible with the provided type, and if so, build and compile a final observable with the specified type.

glopesdev:
To support more flexible programmatic manipulation of expression builder workflows, this PR introduces a new overload to compile an expression builder graph into an observable sequence where the type of the elements in the sequence is known.

This makes it possible to load a workflow and perform a type-safe subscription to access the results of the compiled observable sequence, as shown in the example below:

```c#
using var reader = XmlReader.Create(path);
var workflowBuilder = (WorkflowBuilder)WorkflowBuilder.Serializer.Deserialize(reader);
var observable = workflowBuilder.Workflow.BuildObservable<object>();
```

Both covariance and explicit type conversions are allowed in case of type mismatch between the type of the compiled expression and the expected result type.

Fixes #1915

### DIFF
--- a/Bonsai.Core/Expressions/ExpressionBuilderGraphExtensions.cs
+++ b/Bonsai.Core/Expressions/ExpressionBuilderGraphExtensions.cs
@@ -1041,7 +1041,7 @@ namespace Bonsai.Expressions
             var workflow = source.Build();
             if (!typeof(IObservable<TResult>).IsAssignableFrom(workflow.Type))
             {
-                throw new ArgumentException("The type of the compiled observable sequence is not assignable to the specified sequence type.", nameof(source));
+                workflow = ExpressionBuilder.CoerceMethodArgument(typeof(IObservable<TResult>), workflow);
             }
 
             var observableFactory = Expression.Lambda<Func<IObservable<TResult>>>(workflow).Compile();
--- a/Bonsai.Core/Expressions/ExpressionBuilderGraphExtensions.cs
+++ b/Bonsai.Core/Expressions/ExpressionBuilderGraphExtensions.cs
@@ -1023,6 +1023,31 @@ namespace Bonsai.Expressions
             return observableFactory();
         }
 
+        /// <summary>
+        /// Builds and compiles an expression builder workflow into an observable sequence
+        /// with the specified element type.
+        /// </summary>
+        /// <typeparam name="TResult">The type of the elements in the observable sequence.</typeparam>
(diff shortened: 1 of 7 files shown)


Bonsai.Core/Expressions/ExpressionBuilderGraphExtensions.cs
```diff
@@ -1023,6 +1023,31 @@ public static IObservable<Unit> BuildObservable(this ExpressionBuilderGraph sour
...
+        /// <exception cref="ArgumentException">
+        /// The specified expression builder workflow does not compile into an observable sequence
+        /// with the expected element type.
+        /// </exception>
+        public static IObservable<TResult> BuildObservable<TResult>(this ExpressionBuilderGraph source)
+        {
+            var workflow = source.Build();
+            if (!typeof(IObservable<TResult>).IsAssignableFrom(workflow.Type))
+            {
+                workflow = ExpressionBuilder.CoerceMethodArgument(typeof(IObservable<TResult>), workflow);
```
PathogenDavid, line 1044: Since this method is used for more than just arguments/parameters, would it make sense to rename `CoerceMethodArgument` to something more general like `CoerceType` or `CoerceExpression`?

It doesn't seem to be specific to arguments or argument-like things. (It's also used for properties elsewhere.)

PathogenDavid (approved):
Just one minor thing, good either way.

Bonsai.Core/Expressions/ExpressionBuilderGraphExtensions.cs
glopesdev: Yup, it started like that and then it proved itself too useful not to use 😅 it's fully internal so makes complete sense to rename.

PR #1946, opened 2024-08-02
Exclude build dependencies when counting inputs to new nested node

### LINKED ISSUE #1792: Replacing a node with a group workflow when it receives a visualizer mapping generates a bogus source node
This is something that came up when Gonçalo and I were discussing https://github.com/bonsai-rx/bonsai/issues/1769

Starting with the following workflow: [ReplaceNodeWithGroupWorkflowIssue.zip](https://github.com/bonsai-rx/bonsai/files/15281125/ReplaceNodeWithGroupWorkflowIssue.zip)

ReplaceNodeWithGroupWorkflowIssue

Select the `Untitled` `PublishSubject` node. Replace it with a `GroupWorkflow` node (IE: using the "Replace" option from the context menu in the Toolbox.)

This results in the following within the main workflow:

ReplaceNodeWithGroupWorkflowIssue-Replaced

But within the group workflow you'll see there's a bogus `Source2` (presumably added for the dashed connection for the visualizer mapping.)

image

This source shouldn't be here (and in fact does nothing) as visualizer mappings conceptually apply to the node itself rather than any of its sources.

glopesdev:
Since build dependencies always affect their target node and are not considered proper inputs to operators, they should be excluded when estimating number of inputs to a newly created nested node.

Fixes #1792

### DIFF
--- a/Bonsai.Editor/GraphModel/WorkflowEditor.cs
+++ b/Bonsai.Editor/GraphModel/WorkflowEditor.cs
@@ -1114,9 +1114,10 @@ namespace Bonsai.Editor.GraphModel
             CreateGraphNodeType nodeType)
         {
             // Estimate number of inputs to the nested node
-            var inputCount = workflowBuilder.ArgumentRange.LowerBound;
-            if (nodeType == CreateGraphNodeType.Successor) inputCount = Math.Max(inputCount, selectedNodes.Count());
-            else inputCount = Math.Max(inputCount, selectedNodes.Sum(node => workflow.PredecessorEdges(node).Count()));
+            var inputCount = nodeType == CreateGraphNodeType.Successor
+                ? selectedNodes.Count(node => !node.Value.IsBuildDependency())
+                : selectedNodes.Sum(node => workflow.PredecessorEdges(node).Count(edge => !edge.Item1.Value.IsBuildDependency()));
+            inputCount = Math.Max(workflowBuilder.ArgumentRange.LowerBound, inputCount);
 
             // Limit number of inputs depending on nested operator argument range
             if (!(workflowBuilder is GroupWorkflowBuilder || workflowBuilder.GetType() == typeof(Defer)))
--- a/Bonsai.Core.Tests/TestWorkflow.cs
+++ b/Bonsai.Core.Tests/TestWorkflow.cs
@@ -12,15 +12,21 @@ namespace Bonsai.Core.Tests
         {
         }
 
-        private TestWorkflow(ExpressionBuilderGraph workflow, Node<ExpressionBuilder, ExpressionBuilderArgument> cursor)
+        private TestWorkflow(
+            ExpressionBuilderGraph workflow,
+            Node<ExpressionBuilder, ExpressionBuilderArgument> cursor,
+            int argumentIndex = 0)
         {
             Workflow = workflow ?? throw new ArgumentNullException(nameof(workflow));
+            ArgumentIndex = argumentIndex;
             Cursor = cursor;
         }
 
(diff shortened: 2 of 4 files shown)


PathogenDavid (approved):

Bonsai.Editor/GraphModel/WorkflowEditor.cs
```diff
@@ -1114,9 +1114,10 @@ private void ConfigureBuilder(ExpressionBuilder builder, GraphNode selectedNode,
             CreateGraphNodeType nodeType)
         {
             // Estimate number of inputs to the nested node
-            var inputCount = workflowBuilder.ArgumentRange.LowerBound;
-            if (nodeType == CreateGraphNodeType.Successor) inputCount = Math.Max(inputCount, selectedNodes.Count());
-            else inputCount = Math.Max(inputCount, selectedNodes.Sum(node => workflow.PredecessorEdges(node).Count()));
+            var inputCount = nodeType == CreateGraphNodeType.Successor
+                ? selectedNodes.Count(node => !node.Value.IsBuildDependency())
+                : selectedNodes.Sum(node => workflow.PredecessorEdges(node).Count(edge => !edge.Item1.Value.IsBuildDependency()));
```
PathogenDavid, line 1119: More for my own understanding: Can you explain `IsBuildDependency` here? I expected that the `VisualizerMapping` would *not* be a build dependency and the incoming node *would*, but looking in the debugger (and reading this code) it seems the opposite is true.

I would think `IsBuildDependency` would be true for nodes which are involved in compilation, is that not what it's for? (The documentation on the method isn't very helpful 😅)

PathogenDavid:
How difficult would it be to add a test for this? This fix seems like a good candidate for a test but I know testing stuff in the UI layer can be a pain

Bonsai.Editor/GraphModel/WorkflowEditor.cs
glopesdev: Yeah, `BuildDependency` might not be a great name. Basically `VisualizerMapping` and the other mapping or externalized property nodes are "build dependencies" in the sense that they must be evaluated before their target nodes, but their outputs are not passed directly as input arguments to the `Build` method of their target node.

The reason it is not the best name is that of course the other nodes are also "build dependencies" in the sense that passing arguments also implicitly requires that you build before your target. At the time we didn't want to call them "mapping" nodes because they may play other functions (e.g. externalized property is very different from the others). I think I took this slightly inspired by "development dependencies" in the old NuGet `packages.config` but the choice of name is really not great.

Would be happy to take in proposals to change the name. If you have a better suggestion feel free to open an issue and we can obsolete this one and change it for 3.0.

PathogenDavid: Ah, I can see how the name came into being now. Not sure if a better name immediately comes to mind but I'll let you know if I think of something. It may just be enough to give visualizer mappings and externalized properties as examples in the remarks or something.

glopesdev:
> How difficult would it be to add a test for this? This fix seems like a good candidate for a test but I know testing stuff in the UI layer can be a pain

@PathogenDavid Thanks for the remainder, definitely important to keep doing this, especially for these edge cases, and especially for preparing the migration to a different UI layer.

It turned out to be a bit more difficult than expected, but not because of the UI. The main obstacle was I had to add testing infrastructure for branches, and also had to remember that the initial state of the workflow must be topologically sorted to ensure reversibility of undo/redo operations, so was good to make that explicit at least in a comment.

I think now `TestWorkflow` can be used to build pretty much any workflow programmatically so can almost be considered an API for a scripting editor or DSL.

PR #1949, opened 2024-08-05
Fix `Period` description in `Timer` operator

### LINKED ISSUE #1947: Unexpected Reactive.Timer behavior
Using Bonsai 2.8.5, the behavior on the Timer node appears to be fundamentally different depending on the value of the Period property. 

- If I specify a period of 1 second, the node produces a count value every second.
- If I leave Period set to 00:00:00 a single count of 0 is produced and the sequence then terminates. This does not mesh with the description of the Period parameter.
- If I write a very short period (.e. g 00:00:00.0001000) the timer will produce two values 0 and 1 and then stops.  Again this does not fit with the description of Period

This feels like a race condition to me. e.g. the thread performing checks if counter has passed a due period is using eqaulity operator instead of >=.

bruno-f-cruz:
This PR clarifies/corrects the description of the behavior of `Timer` when the `Period` property is set to 0. With this value, the `Timer` will emit a single event, potentially delayed by `DueTime` and terminate the sequence immediately.

Fixes #1947 documentation gap.

### DIFF
--- a/Bonsai.Core/Reactive/Timer.cs
+++ b/Bonsai.Core/Reactive/Timer.cs
@@ -27,7 +27,7 @@ namespace Bonsai.Reactive
 
         /// <summary>
         /// Gets or sets the period to produce subsequent values. If this value is equal
-        /// to <see cref="TimeSpan.Zero"/> the timer will recur as fast as possible.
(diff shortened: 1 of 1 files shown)


glopesdev:
Closed since we ended up updating functionality of `Timer` in #1957.

PR #1956, opened 2024-08-06
Move layout files to workflow settings folder

### LINKED ISSUE #1955: Reorganize per-workflow settings file structure
## Summary

Move all per-workflow settings to a special folder called `.bonsai/Settings` placed next to the workflow files. Proposed example structure:

```
📦ProjectFolder
 ┣ 📂.bonsai
 ┃ ┣ 📂Settings
 ┃ ┃ ┣ 📜Workflow1.editor
 ┃ ┃ ┣ 📜Workflow1.layout
 ┃ ┃ ┣ 📜Workflow2.editor
 ┃ ┃ ┗ 📜Workflow2.layout
 ┣ 📜Workflow1.bonsai
 ┗ 📜Workflow2.bonsai
```

## Motivation

The work in #1870 has motivated the need to persist additional editor settings on a per-workflow basis (e.g. the explorer tree view state). Saving these passive editor settings in the `.layout` file would be convoluted since these two pieces of state are completely unrelated. Furthermore, it would introduce additional dependencies on the layout file which we are trying to move away from.

Adopting this proposal would help declutter the project folders, and explicitly de-emphasize versioning of `.layout` files. Suggestions for minimizing the impact of this change in current development workflows are presented below.

## Detailed Design

This proposal introduces a new `.bonsai/Settings` folder (not to be confused with the local environment folder) in the project base directory which will be used to store any per-workflow configuration or passive editor settings.

Inside the `.bonsai/Settings` folder, configuration files for each workflow are named using specific extensions depending on the type of recorded configuration state (e.g. `Workflow.layout` instead of the current `Workflow.bonsai.layout`; `Workflow.editor` for the new explorer tree view state).

Using only single dot extensions will allow more convenient retargeting of these configuration files in case the linked workflow file needs to be renamed.

### Backwards compatibility

To facilitate transitioning away from the `.bonsai.layout` files, the editor would keep recognizing existing `.bonsai.layout` files and convert them to the new structure when saving the workflow. We could delete any existing old files when saving the new structure to encourage migration, or otherwise require legacy files to be removed manually.

### Application UI startup

The main drawback of this proposal is that `.bonsai.layout` files are quite often used in application deployment, as they are currently the only option to specify the initial visualizer state for the application at startup time. This is crucial in experiment automation where UI needs to launch with the workflow. Further isolating the `.bonsai.layout` files from the main workflow runs the risk of introducing confusion, especially if we double down on the intent to ignore versioning of these files.

One option to address this drawback is to introduce a new operator for workflows which could be used to explicitly mark which visualizers should be launched and/or made available on startup, as described in #2137. This would help more clearly delineate which visualizers are part of the application, and avoid confusing environment-dependent visualizer window state (e.g. position and location) with application logic.

### Bootstrapper environment folders

The `.bonsai` environment folder is already used as the default name of the portable environment template. To avoid problems in case of overlap, we have chosen to place all configuration files under `.bonsai/Settings` which doesn't yet exist in the bootstrapper structure. This way we can allow environment folders and config folders to coexist in the same base directory.

### Save editor state system-wide or in the bootstrapper folder

Per-workflow editor state could be saved in the bootstrapper folder, next to the `Bonsai.exe.settings` file, either in a folder or files nearby (e.g. similar to the cache files in `Bonsai.exe.WebView2` storing Chromium state for the documentation and annotation control). The internal structure would be similar to the one described in this proposal, and these additional settings files could be saved either per-environment or system-wide in the roaming AppData folder.

Similar to other applications we could compute deterministic pseudo-identifiers from the hash of the absolute file path and use them to identify the corresponding per-workflow config folder.

The main drawback of a centralized location is lack of transparency and making it harder to associate these state files with the workflow itself. Moving or renaming the location of the workflow file would automatically invalidate the corresponding config files and it would not be immediately obvious how to recover the link, especially if hashing is used to identify folders.

### Introduce additional files next to the workflow

Alternatively, we could create additional `.bonsai.*` files next to the workflow, e.g. `Workflow.bonsai.editor`. The drawbacks would be further cluttering of the project folder, harder to rename all config files in one operation, and overall having a more confusing structure for version control systems.

glopesdev:
As detailed in #1955 this PR moves the `.layout` files to the new per-workflow settings folder which aims to declutter project working directories and is designed to be conveniently ignored by version control systems.

New example folder structure for `.layout` files:

```
📦ProjectFolder
 ┣ 📂.bonsai
 ┃ ┣ 📂Workflow1
 ┃ ┃ ┗ 📜.layout
 ┃ ┗ 📂Workflow2
 ┃ ┃ ┗ 📜.layout
 ┣ 📜Workflow1.bonsai
 ┗ 📜Workflow2.bonsai
```

Work on #1870 will allow building on this to save other editor state.

Fixes #1955


glopesdev:
Alternative suggestion from @PathogenDavid if we wanted to avoid potential clashes with environment folder names:

```
📦ProjectFolder
 ┣ 📂.bonsai
 ┃ ┣ 📂Settings
 ┃ ┃ ┣ 📜Workflow1.layout
 ┃ ┃ ┗ 📜Workflow2.layout
 ┣ 📜Workflow1.bonsai
 ┗ 📜Workflow2.bonsai
```

PathogenDavid:
```
📦ProjectFolder
 ┣ 📂.bonsai
 ┃ ┗ 📂Settings
 ┃   ┣ 📜Workflow1.layout
 ┃   ┣ 📜Workflow2.layout
 ┃   ┗ 📂 MyCoolFolder
 ┃     ┗ 📜Workflow3.bonsai
 ┣ 📜Workflow1.bonsai
 ┣ 📜Workflow2.bonsai
 ┗ 📂 MyCoolFolder
   ┗ 📜Workflow3.bonsai
```

PathogenDavid:
During our meeting today we decided to go with the central `.bonsai` approach I demonstrated [here](https://github.com/bonsai-rx/bonsai/pull/1956#issuecomment-2648590381).

Bonsai will prefer any `.bonsai` folder found in the directory tree up from the current workflow, and if none exists it will create one in the same folder as the current workflow.

-------------

Another thing I wanted to mention but didn't want to continue derailing taskboard: Since we want to discourage people from checking `.layout` files into source control, I think we should also generate a self-ignoring `.bonsai/Settings/.gitignore` file that simply contains `*` (which ignores all files and directories under `Settings`, including the `.gitignore` itself.)

We should create this `.gitignore` only if the file doesn't already exist. Anyone who wants to ignore the recommendation and commit them anyway can just replace it with an empty file.

glopesdev:
> Another thing I wanted to mention but didn't want to continue derailing taskboard: Since we want to discourage people from checking `.layout` files into source control, I think we should also generate a self-ignoring `.bonsai/Settings/.gitignore` file that simply contains `*` (which ignores all files and directories under `Settings`, including the `.gitignore` itself.)

@PathogenDavid having this possibility makes sense, but I feel weird baking into the language IDE an assumption about a specific version control system, especially given that Git is not assumed anywhere else in the language / IDE.

We actually even started out the project in Mercurial, and we already had to move everything from hg to git. I know it is hard to imagine right now Git ever losing its dominant position, but still there are other DVCS out there such as Perforce, and I wouldn't want to introduce weird assumptions into possibly valid workflows (we did consider using Perforce ourselves for large asset projects before, so definitely not outside the realm of possibility).

That said, what about we bake this idea into a new project repository template instead?

We did discuss before about best practices for version-controlled projects in Bonsai that would be nice to capture in a template anyway. Because these practices already include bootstrapping an environment in the root `.bonsai` folder, it feels like this structure would already capture whichever workflows are created downstream from the root repository folder quite naturally, so maybe this way we could sidestep the need to bake anything explicitly in the IDE itself?

glopesdev:
Working further on this implementation, I am growing increasingly worried about cache invalidation problems where the `.layout` files may inadvertently become out-of-sync with the main workflow in confusing ways:

- If the workflow file is renamed from the file explorer, the layout loses sync but remains "lost" in the cache
- If some other workflow file is later renamed to match the previously "lost" file, a potentially entirely unrelated layout file may be loaded and lead to arbitrary errors / code execution
- Even without considering file renaming this is a problem. For example in a version-controlled project with a local layout cache in `.gitignore`, simply changing to a different revision will break the association between layout and project file, and lead to confusion
- These issues remain whether we allow searching for shared settings up the path hierarchy or not (the only difference is the probability of conflict / confusion increases the further you move up the hierarchy)

I am not sure how other IDEs such as VS or VS Code deal with such cache problems and VCS for example. I thought of a few possibilities but are not particularly happy with any of them:

1. Run "garbage collection" on the cache every time the IDE loads or a project is opened. Any file settings not matched to an existing workflow file would be deleted. This wouldn't necessarily resolve any of the above issues, but might prevent the probability of confusion by limiting cache file growth. What I find confusing here is you might be inadvertently deleting someone else's layout by simply opening a file with the repo in a different state.
2. Introduce some procedure for renaming workflow files explicitly in the IDE. The renaming procedure would itself take care of moving all workflow cache files. This might improve regular experience when working inside the IDE, but would still not resolve issues when renaming files in the explorer or when changing revisions in the git repo.

glopesdev:
Another possibility is to allow inconsistent cache files to be present, but adopt better standards for robustness against partial and incomplete layout files, as proposed in https://github.com/bonsai-rx/bonsai/issues/2144.

This final option would also mean we wouldn't have to worry about merging this PR as-is and in general might be more future-proof.

glopesdev:
Investigate following edge cases:
  - Access denied or other IO exceptions when recursing up, break if any exception happens
  - Check permissions of the parent folder, could consider breaking search if permissions change

glopesdev:
In https://github.com/bonsai-rx/bonsai/pull/1956/commits/a2fcf3e2d262620e3fa3b867b7ed5f7fe932b1d3 we revised the search strategy to use a centralized settings folder if and only if the workflow file is sitting in a path relative to the parent of the boostrapper folder (i.e. the folder where the executable is launched from).

This is less ambiguous than traversing the full path up to the root, since there are only two possible paths for workflow settings:
  1. the path where the executable is running from (if rooted in a shared path);
  2. the `.bonsai` folder in the local workflow file if no common path is found.

This strategy avoids the above issues with permissions, since it assumes the running executable must be able to write into its own folder. If this is not allowed, other much more serious catastrophic errors will occur either before or after opening the workflow file, since the editor requires read-write access to its own folder during normal operation.

Finally, this strategy also allows more readily for different versions of bonsai to open workflows in remote paths without disrupting their cached development environment, since at most the editor will cache the settings in its own folder, and never in the settings folder of a different environment.

PathogenDavid:
## Git awareness

> but I feel weird baking into the language IDE an assumption about a specific version control system, especially given that Git is not assumed anywhere else in the language / IDE.

I definitely don't love it either, but I'm mainly thinking about creating a pit of success for our users since Git *tends* to be what we push people towards. The best best practices are defaults. (And for anyone not using Git the `.gitignore` would be a sign for them to create their own equivalent.)

I don't feel strongly about this though, just wanted to throw the idea out there.

## Stale layout file problems

> I am not sure how other IDEs such as VS or VS Code deal with such cache problems and VCS for example.

In my experience they best effort ignore the invalid data. Sometimes things barf, but not usually. (C++ IntelliSense still sometimes needs a manual brain refresh.)

> Garbage collection

Not a fan of this solution. Similar to what you alluded to, this would be annoying if you were switching between revisions and like you said only hides the problem.

As for useless layout files sitting around...they're honestly not big enough to worry about IMO.

> Renaming workflows through the IDE

So I don't think this is a good solution to this problem simply because people won't remember to use it (and I don't think they should have to. I've used VCSes that make you do this and it always annoyed me.)

However I do think that this could be a good idea to do as part of the new workflow explorer. (Probably not as part of https://github.com/bonsai-rx/bonsai/pull/1870, but maybe as a feature for it later on.) There's already a node for the root workflow, so it makes sense that you could rename it to rename the file, and this could be nice for child nodes too.

> Store metadata about the workflow file

Ideally we should really just have more robust layout file stuff, but yeah I think this is probably the best short-term solution. I was going to say we got away with not having this for a long time now, but I guess if layout files aren't being committed to Git that mismatch issues are much more likely.

Another idea if you just want a best-effort heuristic thing with minimal changes, you could say that if the workflow modified time is in the future relative to the layout file, then that layout file is invalid. This doesn't catch every scenario, but I think it catches all the typical ones.

-----------------

> Investigate following edge cases: [IO exceptions / write permissions]

We talked about this on Teams but for posterity: I think these edge cases are too weird to worry about and layout files being saved on a best-effort basis is fine.

(But I also think the final solution of using the Bonsai environment only when one's being used is also fine.)

> This could create confusion if we open the workflow file with a bootstrapper located in another location, since it would then skip the centralized settings folder and create a more local one, which would then override the original if/when going back to the original environment.

Not a fan of this edge-case though, but I suppose this just adds more motivation towards improving the experience of double-clicking a `.bonsai` file when there's a local environment in the repo.

glopesdev:
> But I also think the final solution of using the Bonsai environment only when one's being used is also fine.

@PathogenDavid sounds good, let's go with this for now and we can always extend it later if necessary.

> Not a fan of this edge-case though, but I suppose this just adds more motivation towards improving the experience of double-clicking a .bonsai file when there's a local environment in the repo.

This is actually a pretty good idea, might be worth raising an issue for keeping track of it on its own. The current experience is pretty horrible, but maybe since the bootstrapper is already launching subprocesses anyway there could be a way where it just passes the ball to the right environment if one is detected.

PathogenDavid:
> This is actually a pretty good idea, might be worth raising an issue for keeping track of it on its own.

Done! https://github.com/bonsai-rx/bonsai/issues/2148

PR #1957, opened 2024-08-06
Make period a nullable property of reactive timer and clarify documentation

### LINKED ISSUE #1947: Unexpected Reactive.Timer behavior

glopesdev:
Ignoring the value of `TimeSpan.Zero` for the `Period` property is an unfortunate historical legacy. The property should really expose a nullable type when deciding between which `Observable.Timer` overload to call.

This PR changes the type signature of `Period` to be a nullable timespan, but retains the same behavior by requiring `Period` to be larger than zero for periodic timers. We also introduce a slight modification to its serialization behavior to ensure that a value of zero is always serialized as `null`.

This will start the migration away from using `TimeSpan.Zero` as a valid value, and might give us some room to consider reinstating the zero period as a high frequency periodic timer for 3.0.

We could also introduce an automatic conversion at the level of the editor, but this might break embedded include workflows which happen to be using the old timer semantics, since these are not converted automatically by the editor until they are ungrouped and resaved.

The embedded description string was also fixed for the shaders timer.

Fixes #1947

### DIFF
--- a/Bonsai.Shaders/Timer.cs
+++ b/Bonsai.Shaders/Timer.cs
@@ -29,10 +29,10 @@ namespace Bonsai.Shaders
 
         /// <summary>
         /// Gets or sets the period to produce subsequent values. If this value
-        /// is undefined or equal to zero the timer will only fire once.
+        /// is not specified or equal to zero the timer will only fire once.
         /// </summary>
         [XmlIgnore]
-        [Description("The period to produce subsequent values. If this value is equal to zero the timer will recur as fast as possible.")]
+        [Description("The period to produce subsequent values. If this value is not specified or equal to zero the timer will only fire once.")]
         public TimeSpan? Period { get; set; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Bonsai.Shaders/Timer.cs
```diff
@@ -32,7 +32,7 @@ public class Timer : Source<long>
         /// is undefined or equal to zero the timer will only fire once.
         /// </summary>
         [XmlIgnore]
-        [Description("The period to produce subsequent values. If this value is equal to zero the timer will recur as fast as possible.")]
+        [Description("The period to produce subsequent values. If this value is undefined or equal to zero the timer will only fire once.")]
```
PathogenDavid, line 35: Ditto.

PathogenDavid (approved):
Just a nitpick over wording, could go either way

Issue #1948, opened 2024-08-05
Move test projects to modern .NET


glopesdev:
With the move to target .NET 8 we should multi-target all test projects to make sure we run tests across both .NET 4.7.2 and .NET 8.

PR #1958, opened 2024-08-07
Fix test target framework issues

PathogenDavid:
This fixes a mistake made in https://github.com/bonsai-rx/bonsai/pull/1952

* `Bonsai.Editor.Tests` now targets `net8.0-windows` as appropriate
* Made the relevant NuGet warning into an error to avoid this mistake in the future
* Disabled running `net8.0-windows` tests on Linux

The last one was somewhat annoying as `dotnet test` doesn't have a great way to just run tests for all target frameworks except one or something similar. In the end I opted to run `net472`, `net8.0`, and `net8.0-windows` all as discrete steps (which will actually be kinda nice anyway.)

As a result, test projects are only allowed to target one of those three target frameworks. This is enforced by the build system in order to avoid the easy mistake of having an entire test suite skipped without knowing.

-------------

I also looked into why GitHub Actions didn't highlight these warnings in the workflow summary. Turns out it actually [did highlight them](https://github.com/bonsai-rx/bonsai/actions/runs/10257331564), so we're just bad and both missed it. The warnings just didn't show up on the PR because they aren't associated with any particular file within the PR.

It might be nice if warnings like that ended up on the PR anyway. Not sure if there's an easy way to do that unfortunately. Easiest solution is probably that we just get in the habit of reviewing the workflow run as part of code reviews.

As an aside though, I did find out more about how this GitHub Actions feature works. They're called problem matchers and [are documented here](https://github.com/actions/toolkit/blob/f003268b3250d192cf66f306694b34a278011d9b/docs/problem-matchers.md). They can be customized, and it turns out `setup-dotnet` and such tend to add [their own relevant patterns](https://github.com/actions/setup-dotnet/blob/dbebe359e495844f078430a9de2b64d21656c21a/.github/csc.json).

Relates to https://github.com/bonsai-rx/bonsai/issues/1948

### DIFF
--- a/.github/workflows/Bonsai.yml
+++ b/.github/workflows/Bonsai.yml
@@ -101,7 +101,7 @@ jobs:
           release_version: ${{github.event.release.tag_name}}
           workflow_dispatch_version: ${{github.event.inputs.version}}
           workflow_dispatch_will_publish_packages: ${{github.event.inputs.will_publish_packages}}
-   
+
       # ----------------------------------------------------------------------- Build
       - name: Restore
         run: dotnet restore Bonsai.sln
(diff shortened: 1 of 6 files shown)


glopesdev (approved):
Looks great, I didn't know about problem matchers, will have to dig a bit more into it. Would be nice if there was a way to always surface any restore/build warnings to the UI no matter what they are. Catching particular errors is also a good way to go for now.

> As a result, test projects are only allowed to target one of those three target frameworks. This is enforced by the build system in order to avoid the easy mistake of having an entire test suite skipped without knowing.

This is also a good idea, I can definitely imagine this happening when some new .NET comes along, but it's not a bad compromise either.

PR #1969, opened 2024-08-12
Modernize package manager dialog

### LINKED ISSUE #957: Improve experience for embedded license files in package manager
The new standard for NuGet is embedding license files directly in packages to ensure immutability (see [wiki](https://github.com/NuGet/Home/wiki/Packaging-License-within-the-nupkg)). Unfortunately, the package manager experience is not great for these packages in two common scenarios:

1. Packages hosted in local folders will expose the license metadata embedded in the .nupkg but the current dialog only knows how to handle deflated packages, and will crash when trying to access the license terms. A possible workaround is to check whether the package is an instance of `LocalPackageSearchMetadata` and access the local package reader stream for the license file directly.
2. The MyGet server does not correctly handle embedded license files, resulting in both no license metadata and no license URL being available when browsing packages with embedded licenses. This is unfortunate for packages requiring license acceptance as there is currently no way to make the licenses visible.

Unfortunately for 2. it seems that MyGet server version for NuGet has failed to keep up with protocol updates, and it is likely this will continue to be the case. We should probably start considering alternatives for hosting packages in the long term.

### LINKED ISSUE #1603: Add support for filtered package feeds
## Motivation

To ensure a smooth transition from existing community and official MyGet feeds into the NuGet Gallery, we need to introduce built-in support for the nuget.org feed to the package manager, and provide package filter mechanisms to approximate the level of curation currently done by the MyGet feeds.

## Detailed Design

The NuGet v3 API provides several package filter mechanisms allowing constraining existing package source results by using either author, tags or package type filters which can be leveraged by the package manager.

### Proposal 1: Package Filter Checkbox

This option is a straightforward check to enable/disable filtering of packages from any feed, similar to the "Include prerelease" checkbox. As a first pass, this could be implemented by appending package filter tags to the query, e.g. `Bonsai` `Rx`.

### Proposal 2: Feed-specific Filters

As an alternative to, or in addition to, the above filter proposal, we could introduce feed-specific filters, for example to be able to easily curate packages published by the `bonsai` organization for the official "Bonsai Packages" feed.

### Proposal 3: Merge MyGet feeds with NuGet community feed

Migration from the MyGet feed to nuget.org is expected to take several months, if not years. However as space is running out on MyGet we need to start the transition as soon as possible. A useful first step would be to coalesce both the `nuget.org` feed and the `Community Packages` feed into a single curated package source. This way it would be easier to start redirecting developers to publish their packages on nuget.org, and mark the MyGet community feeds as read-only.

### Proposal 4: Mark example packages with a specific package type

Example packages do not by themselves contain any reusable bits, but are used as self-contained applications which are copied on open to a new folder. Their installation and usage is very different from normal packages. At first glance they seem like a good candidate for [setting NuGet package types](https://learn.microsoft.com/en-us/nuget/create-packages/set-package-type) to improve both client behavior and server package search.

Bear in mind that filtering for package types is currently not surfaced (or incorrectly surfaced) via the NuGet client API: https://github.com/NuGet/Home/issues/8915

### Operator Package

- `BonsaiDependency`
- `BonsaiPackage`
- `BonsaiLibrary`
- `BonsaiRx`

### Gallery Package

- `BonsaiGallery`
- `BonsaiExample`
- `BonsaiTemplate`

### LINKED ISSUE #1965: Add a deprecation warning in the Bonsai package manager
You can see a package that is marked as deprecated in nuget, but the only way I know that it is deprecated through the Bonsai UI is if I read its description which doesn't really pop out at me. And that deprecation warning is manually added - not all deprecated packages will have descriptions that indicate as such.

bonsai-deprecated-package

### LINKED ISSUE #2003: Package update button indicates "Package updates are available" even when that statement is false
This button indicates package updates are available even when there aren't actually any package updates available. 

Screenshot 2024-09-06 110523

I suppose it should be greyed out with a different tooltip i.e. "No package updates available" when there are no package updates available?

I like the idea behind this feature a lot btw.

glopesdev:
The package manager dialog has suffered a few changes over the years to support modern NuGet client features. However, this modernization was never finalized and there are still important features of modern NuGet clients which are not included.

This PR updates the styling and functionality of the package manager dialog to display deprecation warnings and latest version strings for all packages inline with the package title, as well as package authors, total package download counts, and package prefix reservation icons. Package operation buttons were replaced with fixed size icons.

A number of changes were required to get the layout to work properly, the most significant of which is adding interactive highlighting to the operation button which is now more important as a call to action. Spacing between package descriptions and various other small layout issues were also fixed.

Operation buttons were included in package details for all actions, including package updates and opening gallery example packages, and support was added for selecting the exact package version.

Finally, a dedicated dialog for viewing embedded license files in local packages was added to resolve #957. This required a refactoring to the internal license acceptance events and dialog. Modern package servers will serve embedded license files and provide the URL in the package search metadata so we will not worry with the remote case.

Filtering for gallery and library packages was added by default to resolve #1603, with a checkbox enabling search for all dependency package types if necessary.

Fixes #2003 
Fixes #1603 
Fixes #1965 
Fixes #957


glopesdev:
Feedback from DCM:
   - If currently installed version is selected, disable the Update / Install button
   - Either improve terminology of "Show dependencies" and / or add tooltip to explain what "dependencies" are or what they are used for
   - Package details link may be broken
   - Consider adding update and version selection buttons to "Installed" tab
   - #2003 
   - Installed tab seems to be using local package metadata so e.g. package deprecation notices are not shown
   - Consider moving to infinite scroll list rather than pages

PR #1989, opened 2024-09-03
Consolidate shared menu icons in global resources

### LINKED ISSUE #1988: Avoid duplicate entries in resource files
Due to legacy serialization rules in designer files, we ended up with duplicate entries in embedded form resources, e.g.:

https://github.com/bonsai-rx/bonsai/blob/b8ece22df2d0669bcc0e1e2661acb77b831194c4/Bonsai.Editor/EditorForm.resx#L133

https://github.com/bonsai-rx/bonsai/blob/b8ece22df2d0669bcc0e1e2661acb77b831194c4/Bonsai.Editor/EditorForm.resx#L399

To avoid this, and because action icons are shared across multiple controls and forms, we should consolidate them into the shared project file.

glopesdev:
This PR moves shared form embedded resources into the global `Resources.resx` file to avoid having the designer create duplicate entries with identical embedded binary image data.

This stems from a long standing quirk of the designer where it will reserialize resources multiple times into different entries in the local embedded resources file instead of referencing, unless images come from the shared project resources. This is done silently and automatically on opening the designer, so it is very easy for the duplication to go unnoticed.

Fixes #1988


Bonsai.Editor/EditorForm.Designer.cs
```diff
@@ -1317,7 +1316,6 @@ private void InitializeComponent()
             // 
             // workflowFileWatcher
             // 
-            this.workflowFileWatcher.EnableRaisingEvents = false;
```
PathogenDavid, line 1320: For posterity: This is the default value for `EnableRaisingEvents`, this is just the designer cleaning up a useless assignment.

PathogenDavid (approved):
LGTM! :shipit:

Bonsai.Editor/EditorForm.Designer.cs
glopesdev: @PathogenDavid thanks for noticing this. I seem to remember having issues with this in the past, so I'll keep an eye out for it. Apparently it was default `true` back in VS 2005 so maybe some quirk of old designer versions.

PR #2016, opened 2024-09-16
Add support for long paths on Windows

glopesdev:
Ever since [Windows 10 version 1607](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later) it has been possible to extend the maximum file path by setting the following registry value:

```reg
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem]
"LongPathsEnabled"=dword:00000001
```

However, applications also need to explicitly opt-in to long path support, which for .NET Framework is done via the `App.manifest` file. This PR explicitly adds the required `longPathAware` element to enable this support.

### DIFF
--- a/Bonsai/App.manifest
+++ b/Bonsai/App.manifest
@@ -48,11 +48,14 @@
   <!-- Indicates that the application is DPI-aware and will not be automatically scaled by Windows at higher
        DPIs. Windows Presentation Foundation (WPF) applications are automatically DPI-aware and do not need 
        to opt in. Windows Forms applications targeting .NET Framework 4.6 that opt into this setting, should 
-       also set the 'EnableWindowsFormsHighDpiAutoResizing' setting to 'true' in their app.config. -->
-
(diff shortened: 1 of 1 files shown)


PathogenDavid (approved):
Thank you for reminding me how awful the long path awareness documentation is.

Not sure what's weirder, the fact that [the manifest requirement isn't mentioned here](https://github.com/microsoft/dotnet/blob/5a5b0e475315c7c8c090bcd942daf4249c758882/Documentation/compatibility/long-path-support.md) or that it *isn't* required for long paths to work in modern .NET (tested locally.)

2025 David here: It's not required because modern .NET [shims the native file functions](https://github.com/dotnet/runtime/blob/12c9ac07ce2cb23823c454754d7627beaa755544/src/libraries/Common/src/Interop/Windows/Kernel32/Interop.CreateFile.cs#L35) to add the per-API `\\?\` opt-in prefix.

PR #2110, opened 2025-01-15
Add MatrixWriter overload for sequences of unmanaged types

### LINKED ISSUE #1578: Consider allowing binary serialization of arbitrary struct types
Currently `MatrixWriter` allows binary serialization of arrays of bytes, but often it would be convenient to allow serializing structures with arbitrary layout (e.g. Point2f, etc). This can possibly be done taking advantage of `Buffer.BlockCopy` similar to the Harp message payload getter:

https://github.com/bonsai-rx/harp/blob/8f79cbd77c360ce734cb3b73ea9252bb4cf9f53a/Bonsai.Harp/HarpMessage.cs#L300

Example overload signature for reference:

```c#
public IObservable<TElement> Process<TElement>(IObservable<TElement> source) where TElement : unmanaged
```

glopesdev:
This PR expands the compatibility of `MatrixWriter` to any arbitrary sequence of [unmanaged types](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/unmanaged-types) by taking advantage of [`Buffer.MemoryCopy`](https://learn.microsoft.com/en-us/dotnet/api/system.buffer.memorycopy?view=net-9.0) to directly marshal the value type into a binary stream.

The new overload signature is:

```c#
public IObservable<TElement> Process<TElement>(IObservable<TElement> source) where TElement : unmanaged
```

`Buffer.MemoryCopy` is compatible across both .NET Framework and .NET 5+ but we might consider later improving the implementation for .NET 5+ by taking advantage of new memory copy performance primitives.

Fixes #1578

### DIFF
--- a/Bonsai.Dsp/MatrixWriter.cs
+++ b/Bonsai.Dsp/MatrixWriter.cs
@@ -72,6 +72,30 @@ namespace Bonsai.Dsp
         }
 
         /// <summary>
+        /// Writes all of the values in an observable sequence to the specified raw binary output stream.
+        /// </summary>
+        /// <typeparam name="TElement">
+        /// The type of the elements in the sequence. This type must be a non-pointer, non-nullable
+        /// unmanaged type.
+        /// </typeparam>
+        /// <param name="source">The sequence of values to write.</param>
+        /// <returns>
+        /// An observable sequence that is identical to the source sequence but where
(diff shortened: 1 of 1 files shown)


PathogenDavid (approved):

jonnew:
yay!!!!

PR #2124, opened 2025-02-03
Avoid launching editor during image export

### LINKED ISSUE #2096: Bonsai `--export-image` can block due to failures from restoring the environment
This happens when generating documentation of https://github.com/bonsai-rx/machinelearning/commit/789b1b891f86d36f328fc9a3bcf83efeb1e2aea5

Bonsai shows the NuGet restore dialog with an error and blocks waiting for the user to interact with the UI to dismiss the error.

This isn't good or useful in a command line context. In a CI context, it causes the run to hang since there's no interactive user to even see the dialog.

The error *is* printed to the terminal already, which is all that's really needed in a context where Bonsai is being run from a script (which is what `--export-image` is for.)

glopesdev:
When the `--export-image` flag was introduced we failed to ensure the editor is not launched during package restore. This PR fixes it by validating the launch editor flag against the state of the export image flag.

Fixes #2096

### DIFF
--- a/Bonsai/Program.cs
+++ b/Bonsai/Program.cs
@@ -86,6 +86,9 @@ namespace Bonsai
             });
             parser.Parse(args);
 
+            if (exportImage)
+                launchEditor = false;
+
             var editorAssembly = typeof(Program).Assembly;
             var editorPath = editorAssembly.Location;
             var editorFolder = Path.GetDirectoryName(editorPath);


Bonsai/Program.cs
```diff
@@ -86,6 +86,7 @@ internal static int Main(string[] args)
             });
             parser.Parse(args);
 
+            launchEditor &= !exportImage;
```
PathogenDavid, line 89: While I do love a nice one-liners, IMO this is more readable at a glance:

```csharp
if (exportImage)
    launchEditor = false;
```

PathogenDavid (approved):
Acceptable as-is but did have a minor suggestion for readability.

Bonsai/Program.cs
glopesdev: @PathogenDavid yeah, I did toggle between the two forms, I was wondering in the future we may have more flags which cancel editor launch but I guess they could all go under the same if statement.

Will come back to this after mulling it over.

Issue #2055, opened 2024-11-12
Extensions cproj is being formated by the editor and stripped of NL


bruno-f-cruz:
Not sure if it is something specific to my machine, but when Bonsai touches the .cproj extensions file, for instance, when a c# transform is added to the workflow with new dependencies, the .cproj file loses all NL characters. This is a bit annoying if one is doing version control on the file.

bruno-f-cruz:
To reproduce:

1 - Open an empty workflow
2 - Add a ScalarBuffer. Save workflow.
3 - Add a CSharpTransform, and save the file in extensions. This will create a csproj and add the OpenCV reference to the file
4 - Repeat 2 and 3, but use CreateBinomial as the source operator which should add the Numerics namespace to csproj.
The file will now be in one line.

![Image](https://github.com/user-attachments/assets/7b270a51-4f59-4d37-8c37-f2972d1234ec)

PathogenDavid:
Thanks a lot for the repro BTW!

Sorry we assumed it was going to be a weird edge case, turns out it was just a dumb regression I introduced back in the first real PR I submitted for Bonsai 😅 https://github.com/bonsai-rx/bonsai/pull/1753

Just submitted a PR to fix this (along with unit tests and some other fixes to issues I found along the way): https://github.com/bonsai-rx/bonsai/pull/2180

PR #2180, opened 2025-03-11
Fix various issues with `Extensions.csproj` handling and add unit testing for it

PathogenDavid:
Was initially focused on https://github.com/bonsai-rx/bonsai/issues/2055, which was caused by an oversight done when refactoring the several redundant XML parsing entrypoints during https://github.com/bonsai-rx/bonsai/pull/1753 (Prior to the refactor, all of them stripped whitespace *except* for the one that loaded the project for adding package references. That exception was lost in the refactor.)

While I was at it I wrote a suite of unit tests related to the handling of `Extensions.csproj` and fixed some other issues I found along the way.

* Fix whitespace being stripped during automatic updates
* Fix fresh `Extensions.csproj` always being written with CRLF newlines
* Fix `ScriptExtensionsProjectMetadata` not actually being immutable as intended
* Fix adding redundant `ItemGroup` when all `PackageReference`s were removed but the group wasn't
* Added automatic exposing of assembly internals to the relevant test assembly
* Upgraded to C# 12 to get raw strings (our `global.json` already requires the .NET 8 SDK anyway)

Fixes https://github.com/bonsai-rx/bonsai/issues/2055

### DIFF
--- a/Bonsai.Configuration.Tests/AssemblyInfo.cs
+++ b/Bonsai.Configuration.Tests/AssemblyInfo.cs
@@ -1,0 +1,3 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+[assembly: Parallelize]
--- a/Bonsai.Configuration.Tests/ScriptExtensionsProjectMetadataTests.cs
+++ b/Bonsai.Configuration.Tests/ScriptExtensionsProjectMetadataTests.cs
@@ -1,0 +1,380 @@
+﻿using System.Linq;
+using System.Xml.Linq;
+using System.Xml.XPath;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Bonsai.Configuration.Tests;
+
+[TestClass]
+public sealed class ScriptExtensionsProjectMetadataTests
+{
+    [TestMethod]
+    [DataRow(true)]
+    [DataRow(false)]
+    public void DefaultValueStateIsValid(bool useNullXDocument)
+    {
+        ScriptExtensionsProjectMetadata metadata = useNullXDocument ? new((XDocument)null) : default;
+        Assert.IsFalse(metadata.Exists);
+        Assert.IsFalse(metadata.AllowUnsafeBlocks);
+        CollectionAssert.DoesNotContain(metadata.GetAssemblyReferences().ToArray(), "System.Windows.Forms.dll"); // Winforms is not enabled by default
+        Assert.AreEqual(0, metadata.GetPackageReferences().Count());
+        Assert.AreEqual
+        (
+            """
+            <Project Sdk="Microsoft.NET.Sdk">            
+
+              <PropertyGroup>
+                <TargetFramework>net472</TargetFramework>
+              </PropertyGroup>
+
+            </Project>
+            """.NormalizeNewlines(),
+            metadata.GetProjectXml()
+        );
+    }
+
+    [TestMethod]
+    [DataRow(true)]
+    [DataRow(false)]
+    public void AllowUnsafeBlocks(bool enabled)
+    {
+        var document = XDocument.Parse
+        ($"""
+        <Project Sdk="Microsoft.NET.Sdk">            
+          <PropertyGroup>
+            <TargetFramework>net472</TargetFramework>
+            <AllowUnsafeBlocks>{enabled}</AllowUnsafeBlocks>
+          </PropertyGroup> 
+        </Project>
+        """, LoadOptions.PreserveWhitespace);
+        var metadata = new ScriptExtensionsProjectMetadata(document);
+        Assert.AreEqual(enabled, metadata.AllowUnsafeBlocks);
+    }
+
+    [TestMethod]
+    [DataRow(true)]
+    [DataRow(false)]
+    public void EnablingWinformsReferencesWinforms(bool enabled)
+    {
+        var document = XDocument.Parse
+        ($"""
+        <Project Sdk="Microsoft.NET.Sdk">            
+          <PropertyGroup>
+            <TargetFramework>net472</TargetFramework>
+            <UseWindowsForms>{enabled}</UseWindowsForms>
+          </PropertyGroup> 
+        </Project>
+        """, LoadOptions.PreserveWhitespace);
+        var metadata = new ScriptExtensionsProjectMetadata(document);
+
+        if (enabled)
+            CollectionAssert.Contains(metadata.GetAssemblyReferences().ToArray(), "System.Windows.Forms.dll");
+        else
+            CollectionAssert.DoesNotContain(metadata.GetAssemblyReferences().ToArray(), "System.Windows.Forms.dll");
+    }
+
+    [TestMethod]
+    public void GetPackageReferences()
+    {
+        var document = XDocument.Parse
+        ($"""
+        <Project Sdk="Microsoft.NET.Sdk">            
+          <PropertyGroup>
+            <TargetFramework>net472</TargetFramework>
+          </PropertyGroup> 
+          <ItemGroup>
+            <PackageReference Include="FakePackageA" Version="3226.42.1337" />
+            <PackageReference Include="FakePackageB" Version="126.484.30" />
+          </ItemGroup>
+        </Project>
+        """, LoadOptions.PreserveWhitespace);
+        var metadata = new ScriptExtensionsProjectMetadata(document);
+        var references = metadata.GetPackageReferences();
+        CollectionAssert.AreEquivalent((string[])["FakePackageA", "FakePackageB"], references.ToArray());
+    }
+
+    [TestMethod]
+    public void GetPackageReferencesTwoItemGroups()
+    {
+        var document = XDocument.Parse
+        ($"""
+        <Project Sdk="Microsoft.NET.Sdk">            
+          <PropertyGroup>
+            <TargetFramework>net472</TargetFramework>
+          </PropertyGroup> 
+          <ItemGroup>
+            <PackageReference Include="FakePackageA" Version="3226.42.1337" />
+          </ItemGroup>
+          <ItemGroup>
+            <PackageReference Include="FakePackageB" Version="126.484.30" />
+          </ItemGroup>
+        </Project>
+        """, LoadOptions.PreserveWhitespace);
+        var metadata = new ScriptExtensionsProjectMetadata(document);
+        var references = metadata.GetPackageReferences();
+        CollectionAssert.AreEquivalent((string[])["FakePackageA", "FakePackageB"], references.ToArray());
+    }
+
+    [TestMethod]
+    public void AddNewPackageReference()
+    {
+        var document = XDocument.Parse
+        ($"""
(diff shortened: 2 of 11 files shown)


Bonsai.Configuration.Tests/ScriptExtensionsProjectMetadataTests.cs
```diff
@@ -0,0 +1,380 @@
...
+        metadata = metadata.AddPackageReferences([new PackageReference("FakePackageA", "3226.42.1336")]);
+        var xml = metadata.GetProjectXml();
+        var xElement = XElement.Parse(xml);
+        var itemGroups = xElement.Descendants("ItemGroup").ToArray();
+        var packageReferences = xElement.Descendants("PackageReference").ToArray();
+
+        CollectionAssert.AreEquivalent((string[])["FakePackageA", "FakePackageB"], metadata.GetPackageReferences().ToArray());
+
+        // Package was not updated
+        //TODO: This is how the implementation is written, but is this behavior correct?
```
PathogenDavid, line 366: This test covers the behavior provided by this check:

https://github.com/bonsai-rx/bonsai/blob/5279025d323e4e7f01a7de373fca4627206a0430/Bonsai.Configuration/ScriptExtensionsProjectMetadata.cs#L98

It's been like this since 2018 [when automatic `Extensions.csproj` updating was introduced](https://github.com/bonsai-rx/bonsai/commit/e1ba312c6950bff92101f39469b86f2ee427d308#diff-506e74263843b950cb380fa7720d14254eb9fe7cd60b3941dd8a06494a8aabc5R73).

Is this behavior actually a good idea? It seems to me like the `Extensions.csproj` should always match the Bonsai environment since the two not matching only causes confusion.

PathogenDavid:
Do we want to special-case saving to automatically re-add newlines if the original project had none in order to undo the damage caused by https://github.com/bonsai-rx/bonsai/issues/2055 ?

Should be as simple as skipping `SaveOptions.DisableFormatting` if the first descendant of the `Root` is an `XText` not containing a `\n`, but on the other hand anyone who cares can just ask their IDE to reformat the file to restore sensible whitespace.

glopesdev:
> Do we want to special-case saving to automatically re-add newlines if the original project had none in order to undo the damage caused by https://github.com/bonsai-rx/bonsai/issues/2055?

This is probably a good idea, especially if we do it together with the check on the `Extensions.csproj` at load-time for mismatched versions. That check could include this verification and could also trigger the option to resynch the project.

Bonsai.sln
glopesdev, line 1: I was curious about why so many project configurations got removed here. They don't seem to be added elsewhere, so just wondering if anything got dropped in the process.

tooling/Common.csproj.targets
glopesdev, line 1: Interesting, I wasn't aware you could specify internals visible to in MS Build props, I guess that works too. Out of curiosity, does the attribute still get added to the published assembly metadata, or is this somehow constrained to the build process itself?

glopesdev (commented):
Looks good, the new test infrastructure is a really nice touch. I added minor comments / questions and just one typo I found.

Bonsai.Configuration/ScriptExtensionsProjectMetadata.cs
```diff
@@ -24,20 +25,27 @@ public readonly struct ScriptExtensionsProjectMetadata
         const string UseWindowsFormsElement = "UseWindowsForms";
         const string AllowUnsafeBlocksElement = "AllowUnsafeBlocks";
 
-        const string ProjectFileTemplate = @"<Project Sdk=""Microsoft.NET.Sdk"">
+        private static XElement ProjectFileTemplate
+            => XElement.Parse
+            ("""
```
glopesdev, line 30: One last question about raw string literals: the new lines here will be whatever is in source control I guess when building the package, and either option will work for loading.

When saving the file to disk, will the serializer adopt file system conventions when writing the `XElement`?

Bonsai.sln
PathogenDavid: Good catch.

These used to be the configuration entries for `Bonsai.Scripting`/`.IronPython`/`.Design`. They should've been removed when those projects were removed but [got left behind somehow](https://github.com/bonsai-rx/bonsai/commit/e9e00ba1cfc4a6ae6b4bf9c5d0ebd033fc7bc8ff#diff-28d92b6d20fe9404d05217e0e9d61c3baa965e003863f249b7906313c38b89c7).

Visual Studio just automatically cleaned them up when I added the project and I didn't notice. (I think I probably thought it was a reorder rather than a removal since that happens sometimes too.)

tooling/Common.csproj.targets
PathogenDavid: Nope, no magic here. All this does is synthesize the attribute into `<project-name>.AssemblyInfo.cs` in the `obj` folder (similar to all the assembly metadata attributes populated from NuGet metadata.)

The main advantage here is that it makes exposing internals to tests automatic since it's typically what you want. (Unless you're a zealot about only testing what's accessible via the public API surface.)

PathogenDavid: It's also worth pointing out that any assembly-level attribute can be added this way. These items are just sugar for the `AssemblyAttribute` MSBuild item, which you can see [in the implementation](https://github.com/dotnet/sdk/blob/7eda1e985ae8a2e328ff7b95100ba003c876cc95/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets#L119-L123).

Bonsai.Configuration.Tests/AssemblyInfo.cs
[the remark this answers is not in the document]
PathogenDavid: I actually almost added it to be synthesized in `Common.csproj.targets`, but I didn't want to assume all of the existing tests would be happy running concurrently. This was also when we discussed possibly moving to xUnit, which felt like an opportunity to kill two birds with one stone.

Definitely something we should do, but let's maybe do it after 2.9 since dealing with flaky tests near a release would be a pain and it's not especially urgent. I created an issue to follow up: https://github.com/bonsai-rx/bonsai/issues/2274

Bonsai.Configuration.Tests/ScriptExtensionsProjectMetadataTests.cs
PathogenDavid: Created an issue to follow up on this: https://github.com/bonsai-rx/bonsai/issues/2275

Bonsai.Configuration/ScriptExtensionsProjectMetadata.cs
PathogenDavid: > When saving the file to disk, will the serializer adopt file system conventions when writing the `XElement`?

Yes, despite the `LoadOptions.PreserveWhitespace` and `SaveOptions.DisableFormatting`, the newlines will be normalized. ([.NET Lab](https://lab.razor.fyi/#XY_BSsNAEIaRetA9iU8w5pSAbB8gbRGsolKwNIUW6cE1GevCZifuTiqiHn0In8irB9_DNxBJ05TW0-z88_3_zIrflhBDR3Oncpn6w59W6bWdQ_LsGfNYbHZymhs50PYxFmKMnsOgc4HGUG9mAToTciaDdm9mO-1aDqL4H-e2SbfNCs-KdQoL0hksjZ5dtZ4xL4xinOYmEi8CAGB6ZjBHy-sZdNeaHCrnMdxwHcOAVHZdsCbr5dChR7fAyYNm9IVKEV63gAR5oC1e2nuK4u11joihCxaf1vkN0qe0XDJZ86i5yrJiVv-hkouyGjekHBHxkRxTsgTCRC2wOaavvbozeE4uV8zazldZp2Q9GZQTpxmrc8M6Vo6wMCrFMJi54BhOqhJtiLYWbRBFsXgTVy1X2pv9va_P949vPNi93fkD))

The [XML spec actually mandates](https://www.w3.org/TR/REC-xml/#sec-line-ends) that conforming parsers normalize newlines.

(The [implementation](https://github.com/dotnet/runtime/blob/6a3747bf868f5a7774c336103c0b48e7275958e5/src/libraries/System.Private.Xml/src/System/Xml/Core/XmlTextReaderImpl.cs#L5566-L5591) technically allows for disabling this normalization, but if you follow the breadcrumbs it only exists to support the long-deprecated `XmlTextReader` class, which [allowed disabling it](https://learn.microsoft.com/en-us/dotnet/api/system.xml.xmltextreader.normalization?view=net-8.0#system-xml-xmltextreader-normalization).)

Bonsai.Configuration.Tests/ScriptExtensionsTests.cs
```diff
@@ -0,0 +1,211 @@
...
+        var packageConfiguration = CreateDummyPackageConfiguration();
+        using var extensions = new ScriptExtensions(packageConfiguration, $"{nameof(WillLoadProjectMetadata)}Output");
+        ScriptExtensionsProjectMetadata metadata = extensions.LoadProjectMetadata();
+        Assert.IsTrue(metadata.Exists);
+        Assert.IsTrue(metadata.AllowUnsafeBlocks);
+        CollectionAssert.Contains(metadata.GetPackageReferences().ToArray(), "FakePackage");
+    }
+
+    [TestMethod]
+    public void UpdateProjectMetdataInitializesProject()
```
glopesdev, line 98:
```suggestion
    public void UpdateProjectMetadataInitializesProject()
```

There were actually two minor typos in this method name and only one of them was fixed, sorry I didn't make that clear in the first suggestion.

tooling/Common.csproj.targets
glopesdev: > The main advantage here is that it makes exposing internals to tests automatic since it's typically what you want. (Unless you're a zealot about only testing what's accessible via the public API surface.)

I see, actually I think in general I do prefer testing only through the public API surface since I find it makes the code less coupled (i.e. makes internal implementation really free to vary as long as public tests all pass). In that case I might prefer to actually apply this on a case-by-case basis.

I was checking this case and it seems all classes are public anyway, was there a case in which you needed access to an internal member for testing?

Bonsai.Configuration.Tests/ScriptExtensionsTests.cs
PathogenDavid: Fixed! My bad for not double-checking 😅

tooling/Common.csproj.targets
PathogenDavid: Responded on Teams, but for posterity:

I definitely agree that you *shouldn't* touch public APIs just for fun, but I think it's inevitable that unit tests will need to poke at internals to test certain things. If you start relying on internals unnecessarily only because of the presence of `InternalsVisibleTo`, you were already being sloppy.

The tests in this PR need it to access the [`ScriptExtensionsProjectMetadata` constructor](https://github.com/NgrDavid/bonsai/blob/782598d3b2b531ead1b3b8b64f6b7e5e86d892ca/Bonsai.Configuration/ScriptExtensionsProjectMetadata.cs#L40), which should only really be created by `ScriptExtensions`.

glopesdev (approved):

PR #2201, opened 2025-03-27
Ensure property mapping and member selector operators prefer derived properties

### LINKED ISSUE #2200: Property mapping on same name property in derived class results in build error
Following #2192 property mapping on `Value` in the `StringProperty` raises a build exception. This is due to the implementation of the editor declaration using a same-name property which hides the base class property, resulting in an ambiguous match when calling `Expression.Property`:

https://github.com/bonsai-rx/bonsai/blob/25a76d28bee0937952e987db119bb9fec6d292a9/Bonsai.Core/Expressions/ExpressionBuilder.cs#L1104

While the regression is caused by #2192, this is probably an error in the language compiler, since by definition the derived property should be preferred in this case.

glopesdev:
By default, the named overload to the [`Expression.Property`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.expressions.expression.property?view=net-9.0#system-linq-expressions-expression-property(system-linq-expressions-expression-system-string)) and [`Expression.PropertyOrField`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.expressions.expression.propertyorfield?view=net-9.0) methods may throw an exception when accessing a property or field that hides an inherited member with the same name, due to the default rules used by the underlying [`System.Type.GetProperty`](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-type-getproperty) method where an `AmbiguousMatchException` is thrown.

Here we rewrite several uses of these named accessors to ensure the most derived members with the same name are always preferred by searching incrementally through the type hierarchy.

Fixes #2200

### DIFF
--- a/Bonsai.Core.Tests/ExpressionBuilderGraphTests.cs
+++ b/Bonsai.Core.Tests/ExpressionBuilderGraphTests.cs
@@ -300,6 +300,17 @@ namespace Bonsai.Core.Tests
         }
 
         [TestMethod]
+        [ExpectedException(typeof(WorkflowBuildException))]
+        public void Build_PropertyMappingToMissingProperty_ThrowsWorkflowBuildException()
+        {
+            new TestWorkflow()
+                .AppendValue(1)
+                .AppendPropertyMapping(nameof(DerivedNewProperty.AnotherValue))
+                .AppendCombinator(new DerivedValueProperty())
+                .BuildObservable<Unit>();
+        }
+
+        [TestMethod]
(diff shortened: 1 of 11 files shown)


glopesdev:
As a note for future performance investigations: looking into the source code of the reflection system this incremental search did not seem overly expensive as there is a zero-allocation cache for finding named property and field candidates. Any questions regarding performance of this approach should be benchmarked first to determine whether this is the true bottleneck.

PR #2205, opened 2025-03-27
Expose texture array interface on texture sequence

### LINKED ISSUE #1968: Consider exposing the internal frame count in ImageSequence
Relates to the Bonsai.Shaders package.

In some cases, it is useful to be able to access the total frame count of a texture sequence at runtime in Bonsai. For example, when scrubbing through a texture sequence one may want to 'wrap' back around after the maximum or minimum index of the sequence has been reached. To do this, the limits of the texture sequence indices must be available.

As far as I can tell it's very tricky to access these values either through current Bonsai operators or extensions. [ImageSequence](https://github.com/bonsai-rx/bonsai/blob/main/Bonsai.Shaders/Configuration/ImageSequence.cs) can access this internally via `GetFrames` but the public property `FrameCount` is not updated if there is no user provided value (though the internal frame count can be overwritten if there is [user-provided value for `FrameCount`](https://github.com/bonsai-rx/bonsai/blob/70520d90f48d5fcf22de64bd7db5282cebe5657d/Bonsai.Shaders/Configuration/ImageSequence.cs#L111)). A simple solution might be to update the `FrameCount` property during the `GetFrames` method.

glopesdev:
As discussed in #1968 and #1999, there is often a need to explicitly access the array of textures stored in a texture sequence for custom purposes. Here we provide a new interface `ITextureArray` matching the already public class `TextureArray` and also provide an adapter to this interface via the `TextureSequence` class.

This allows downcasting from a texture sequence into `ITextureArray` for the purposes of accessing the internal texture handles:

```c#
if (window.ResourceManager.Load<Texture>(name) is ITextureArray array)
{
    for (int i = 0; i < array.Length; i++)
        GL.BindTexture(TextureTarget, array[i]);
}
```

Fixes #1968

### DIFF
--- a/Bonsai.Shaders/ITextureArray.cs
+++ b/Bonsai.Shaders/ITextureArray.cs
@@ -1,0 +1,26 @@
+﻿using System.Collections.Generic;
+
+namespace Bonsai.Shaders
+{
+    /// <summary>
+    /// Represents a read-only array of texture objects that can be accessed by index.
+    /// </summary>
+    public interface ITextureArray : IEnumerable<int>
+    {
+        /// <summary>
+        /// Gets the total number of texture objects in the array.
+        /// </summary>
+        int Length { get; }
+
+        /// <summary>
+        /// Gets the texture object at the specified index.
+        /// </summary>
+        /// <param name="index">
+        /// The zero-based index of the texture object to get.
+        /// </param>
+        /// <returns>
+        /// The handle to the texture object at the specified index.
+        /// </returns>
+        int this[int index] { get; }
+    }
+}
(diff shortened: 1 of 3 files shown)


RoboDoig (approved):
In my case, I need to be able to extract the number of textures in a sequence at runtime with something like:
```
public override IObservable<int> Generate()
    {
        return ShaderManager.WindowSource.Select(window => {
            var texture = window.ResourceManager.Load<Texture>(TextureName);

            if (texture is ITextureArray) {
                ITextureArray tex = texture as ITextureArray;
                return tex.Count();
            } else {
                return 0;
            }
        });
    }
```

Tested this PR with that extension operator and all seems to be working correctly.

PR #2206, opened 2025-03-27
Deprecate obsolete visualizer operators

### LINKED ISSUE #1863: Consider deprecating `Visualizers` package
A few of the classes of the `Visualizer`'s package have now been moved to the new `GUI.ZedGraph`. Since the support for UI is so recent in Bonsai, we should consider deprecating the `Visualizers` package to prevent confusing people as to which to use. Alternatively, we could make the `Visualizers` package a "meta-package", similar to `Bonsai.Neuro` or `Bonsai.Expressions`, that installs GUI and GUI.ZedGraph.

glopesdev:
With the release of the new [bonsai-rx/gui](https://github.com/bonsai-rx/gui) package, several operators once present in `Bonsai.Design.Visualizers` became obsolete.

While the design visualizers package also incorporates some built-in implicit visualizers, all explicit visualization and GUI building operator development is expected to continue over at [bonsai-rx/gui](https://github.com/bonsai-rx/gui) so to avoid confusion we deprecate all operators whose development has been transferred.

Fixes #1863


bruno-f-cruz (approved):
Nice to see this in 2.9!

Slightly tangencial, it seems the attribute supports a string overload. Does this information surface to the user in the editor? An interesting possibility would be to encourage people to replace this namespace with the GUI one.

glopesdev:
@bruno-f-cruz good point, I am not sure if it does get surfaced, but probably it should.

That might be a good thing to a raise as a separate issue in and of itself.

glopesdev:
> Slightly tangencial, it seems the attribute supports a string overload. Does this information surface to the user in the editor?

@bruno-f-cruz this is resolved in #2208

PR #2213, opened 2025-03-29
Avoid loading visualizer layout on mismatched type name

### LINKED ISSUE #1340: Mismatched layout file terminates workflow with error when using 'visualizermapping'
Not fully sure how to reproduce this bug yet.
A combination of an older layout file with a recent workflow using `visualizermapping` seems to induce the following error when attempting to start the workflow:
image
The issue seems to be fixed by deleting the bonsai.layout file and reloading the workflow.
Deleting the node without reloading the workflow does not fix the issue either.

In the future it might be worth implementing a fail-safe routine to prevent these deadlocks. Acknowledging that the workflow cannot automatically recover the layout state, and simply reset it to allow the user to run the workflow at the cost of having to create a new layout from scratch.

glopesdev:
The editor used to perform only light touch validation of layout file contents. Specifically any visualizer type names would not be matched against the types actually supported by the target operator. This would cause visualizer initialization to fail at start time, effectively preventing the workflow from running. The issue would remain until the layout file is either deleted or corrected, either of which might be hard to diagnose without insight into editor operation details.

To avoid this complex situation, we introduce here additional validation that checks at load time whether the type names in the layout file can be applied to their targets. If not, the entire layout restore step fails, and the user is prompted with a message to delete the layout file.

Fixes #1340


bruno-f-cruz (commented):
Didn't look too closely, but it seems like this runs regardless of whether the layout is passed explicitly via the cli, or even if it runs under --no-editor. Does it make sense to still prompt the user under those cases? Otherwise lgtm!

glopesdev:
@bruno-f-cruz the check is indeed performed in all cases, but the offer to delete is only shown from inside the editor form.

It is a good point though that we should decide whether it is a bug or a feature that mismatched layouts cause CLI calls to throw an error.

PR #2239, opened 2025-04-14
Disable confusing editor actions on disabled nodes

### LINKED ISSUE #2238: Disallow launching visualizers from disabled nodes
Due mostly to historical reasons, it is currently possible to launch a visualizer by double-clicking on a disabled node.

Although disabled nodes will correctly propagate the visualizer element from the input source, if any, this behavior is potentially confusing for a language element which is akin to commented out code.

glopesdev:
Disabled nodes should be treated akin to commented out code. Context menu actions and other editor actions other than branching should be disallowed to prevent accidental creation of confusing workflows.

Fixes #2238

### DIFF
--- a/Bonsai.Editor/GraphView/WorkflowGraphView.cs
+++ b/Bonsai.Editor/GraphView/WorkflowGraphView.cs
@@ -1603,13 +1603,13 @@ namespace Bonsai.Editor.GraphView
                 }
 
                 var builder = WorkflowEditor.GetGraphNodeBuilder(selectedNode);
-                defaultEditorToolStripMenuItem.Enabled = HasDefaultEditor(builder);
-                goToDefinitionToolStripMenuItem.Enabled = HasDefinition(builder);
-                docsToolStripMenuItem.Enabled = true;
-
-                var workflowElement = ExpressionBuilder.GetWorkflowElement(builder);
-                if (workflowElement != null)
+                if (builder is not DisableBuilder)
                 {
+                    defaultEditorToolStripMenuItem.Enabled = HasDefaultEditor(builder);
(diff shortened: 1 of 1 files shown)


PathogenDavid:
@glopesdev Should the output item be disabled as well? It *technically* works and I can see why it works the way it does, but I think the behavior is slightly confusing and if the user really wanted to select properties of the value before the disabled node they should just right-click the node before it.

Node enabled (shows output menu for node's output):

image

Node disabled (output menu is really for preceding node's output):

image

(AFAIK this menu item is also the main/only way to see the return type of the node, so the current behavior removes you ability to inspect that.)

glopesdev:
@PathogenDavid In that case I think the behavior does make sense for a few reasons. The way to see it is that you are indeed inspecting the behavior of the sequence at that point, and all the structural manipulations you create are functional and expected.

I will leave below a table of various cases of building a workflow around a disabled node. We can use them and their labels to discuss allowed / disallowed operations in both cases.

workflow | label |
| ---------  | ------ |
| image | Original |
| image | DisableGray |
|image | DeleteThresh |
| image | BranchSource |
| image | BranchDisabled |
| image | BranchGray |

If we disallow selecting outputs then the transitions `DisableGray --> BranchDisabled` and `DeleteThresh --> DisableGray` will not be available. We would be left only with the option of branching from the source.

Essentially it feels to me unnatural that we try to force pruning the entire tree from growing at a disabled leaf, while at the same time allowing to disable a node at any point in the tree.

PathogenDavid:
Apparently I forgot to respond to this. While I don't love it, I definitely agree it's necessary to enable the branching situations you described.

PR #2240, opened 2025-04-17
Add new VisualizerWindow operator

### LINKED ISSUE #2137: Allow embedding UI startup logic in the workflow
## Summary

Introduce a new operator for workflows (e.g. `ShowVisualizer`) which can be used to explicitly mark which visualizers should be launched and/or made available on startup.

### Application UI startup

The `.bonsai.layout` files are quite often used in application deployment, as they are currently the only option available to specify the initial visualizer state for the application at startup time. This is crucial in experiment automation where UI needs to launch with the workflow. As first discussed in #1955, further isolating the `.bonsai.layout` files from the main workflow runs the risk of introducing confusion, especially if we double down on the intent to ignore versioning of these files altogether.

One option to address this drawback is to introduce a new operator for workflows (e.g. `ShowVisualizer`) and explicitly mark which visualizers should be launched and/or made available on startup. This could also be leveraged as a starting point to avoid confusing environment-dependent visualizer window state (e.g. position and location) with application logic.

Explicitly launched debugger visualizers would continue to save their environment-specific state in the `.layout` file.

## Detailed Design

TBD

## Unresolved Questions

- Flesh out details of the new launch visualizer operator
  - Name: `ShowVisualizer`, `LaunchVisualizer`, `Visualize`, `VisualizerLauncher`
  - Namespace: `Bonsai.Design`
  - Proposed Properties:
    - `Visible`
    - `DefaultBounds`
    - `DefaultWindowState` (borders?)
    - `Title` / `Text` (or reuse externalize property for `Name` from table layout)

glopesdev:
This operator allows declaring which specific visualizer windows should be launched on workflow start, and makes it possible to have frontend applications launch correctly even in the absence of a visualizer layout file.

The new `VisualizerWindow` operator does not allow changing the visualizer type at runtime, but it is possible to relaunch the window from the context menu in case it is closed.

Note that while basic window settings may be cached in the local layout file, runtime visualizer settings are not currently saved for explicitly launched visualizers. If required, such logic should be embedded in the frontend UI itself.

`VisualizerWindow` properties which are explicitly set will always overrule the cache.

Fixes #2137

### DIFF
--- a/Bonsai.Core/Expressions/ExpressionBuilderGraphExtensions.cs
+++ b/Bonsai.Core/Expressions/ExpressionBuilderGraphExtensions.cs
@@ -60,42 +60,29 @@ namespace Bonsai.Expressions
                 throw new ArgumentException("The workflow property name cannot be null or whitespace.", nameof(name));
             }
 
-            object component = source;
             var memberChain = name.Split(new[] { ExpressionHelper.MemberSeparator }, StringSplitOptions.RemoveEmptyEntries);
             for (int i = 0; i < memberChain.Length - 1; i++)
             {
-                var namedBuilders = (from node in source
-                                     let builder = ExpressionBuilder.Unwrap(node.Value) as ISupportPropertyAssignment
-                                     where builder != null && builder.Name == memberChain[i]
-                                     select builder).ToArray();
-                if (namedBuilders.Length == 0)
+                var workflowBuilders = (from node in source
+                                        let builder = ExpressionBuilder.Unwrap(node.Value) as WorkflowExpressionBuilder
+                                        where builder != null && builder.Name == memberChain[i]
+                                        select builder).ToArray();
+                if (workflowBuilders.Length == 0)
                 {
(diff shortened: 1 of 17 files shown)


glopesdev:
@PathogenDavid In commit https://github.com/bonsai-rx/bonsai/pull/2240/commits/3c1f56ebaed19e3c443593ebac8fd69637db5a17 we have reverted a big part of #2263 which would allow directly setting properties in the `VisualizerWindow` operator.

In conversations with @bruno-f-cruz and @jfrazao we revisited the issue of externalizing compile-time properties and how this has precedents, e.g. subject names and parse patterns, in other nodes. There is currently no mechanism available to indicate the purpose of a property (that is up to the code of the operator), and so to address this problem consistently we would need a wider discussion, and most likely we will only be able to address this for 3.0.

Therefore for now we have decided to allow externalizing properties in `VisualizerWindow` in the usual way. To configure these properties from the CLI we will need to make them "public" by externalizing them to the top-level, or to a named group workflow at the top-level.

PR #2257, opened 2025-04-26
Ensure await of background tasks

### LINKED ISSUE #2256: Unhandled exceptions during asynchronous workflow opening are ignored
Due to conversion of all background event handling to fire-and-forget tasks, any unhandled exceptions are ignored. Although all tasks receive a cancellation token tied to form closure, this is not the same as setting up a proper awaiter which will tie task termination to the top-level message loop.

There seems to be two options: either configure an awaiter for all background tasks, or revert the change and move back to reactive subscriptions which are cancelled on form closure. In the latter case, it is possible we can keep async loading of workflows still, as that task is not long-running and fits well within the `async` / `await` framework.

glopesdev:
Tasks started without an awaiter will not propagate exceptions to the top-level form. Here we explore setting up an explicit awaiter for all background tasks to ensure both error propagation and cancellation.

Fixes #2256

### DIFF
--- a/Bonsai.Editor/EditorForm.cs
+++ b/Bonsai.Editor/EditorForm.cs
@@ -289,13 +289,20 @@ namespace Bonsai.Editor
             formCancellation.Cancel();
         }
 
-        void HandleUpdatesAvailable(bool value)
+        Task HandleUpdatesAvailableAsync(CancellationToken cancellationToken)
         {
-            if (value) toolStrip.Items.Add(statusUpdateAvailableLabel);
-            else toolStrip.Items.Remove(statusUpdateAvailableLabel);
+            return updatesAvailable
+                .ObserveOn(formScheduler)
+                .Do(value =>
+                {
+                    if (value) toolStrip.Items.Add(statusUpdateAvailableLabel);
+                    else toolStrip.Items.Remove(statusUpdateAvailableLabel);
+                })
+                .IgnoreElements()
+                .ToTask(cancellationToken);
         }
 
-        protected override void OnLoad(EventArgs e)
+        protected override async void OnLoad(EventArgs e)
         {
             RestoreEditorSettings();
             var initialFileName = FileName;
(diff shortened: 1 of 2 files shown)


glopesdev:
The main consideration here is how to model background event handlers constructed using Rx operators Each background task needs to be cancelled on form closure, and bubble up exceptions to the top-level UI.

The async version converts each observable into a `Task` and waits for all them to complete from `Form.Load`:

```c#
var backgroundTasks = Task.WhenAny(Task.WhenAll(
    InitializeSubjectSourcesAsync(formCancellation.Token),
    InitializeWorkflowFileWatcherAsync(formCancellation.Token),
    InitializeWorkflowExplorerWatcherAsync(formCancellation.Token),
    updatesAvailable.ObserveOn(formScheduler).Do(HandleUpdatesAvailable).ToTask(formCancellation.Token)),
    formCancellation.Token.ToTask());
await backgroundTasks;
```

This is relatively clean, although `Task.WhenAll` sadly does not itself support cancellation, which means we need to convert the token itself into a `Task` and implement cancellation with `Task.WhenAny`. Exceptions are routed corrrectly and closing the form unsubscribes from all event handlers cleanly.

Alternatively, we can leave everything as `IObservable`, combine everything with `Observable.Merge` and register the subscription cancellation action on the token:

```c#
var backgroundTasks = Observable.Merge(
    InitializeSubjectSourcesAsync(),
    InitializeWorkflowFileWatcherAsync(),
    InitializeWorkflowExplorerWatcherAsync(),
    updatesAvailable.ObserveOn(formScheduler).Do(HandleUpdatesAvailable).Select(_ => Unit.Default))
    .Subscribe();
formCancellation.Token.Register(backgroundTasks.Dispose);
```

None of the patterns is perfect syntax-wise, but the latter pattern at least does not require any substantial wrangling to interact with the cancellation token.

glopesdev:
Closing in favor of #2258

PR #2267, opened 2025-05-03
Ensure workflow can be built on image export

### LINKED ISSUE #1892: Consider printing a command line warning when exporting SVG's if dependencies are not present
When automatically generating an SVG of a workflow, using `--export-image` from the command line, there is no warning currently if a dependency for the given workflow is not loaded. Instead, the resulting SVG includes a node with cross-hatches.

It would be useful to know that an image would be missing node information, but specifically this would be useful for automating the export of workflows to create documentation. Having some way to check that the command worked instead of silently generating the image with X-ed out nodes would be beneficial in GitHub Actions workflows to point out when manual intervention might be necessary.

This seems tangentially related to #1725 as well.

glopesdev:
Deserializing a workflow may introduce placeholder elements for any unknown types so the workflow may be shown in the visual editor. Even if there are no placeholder elements, the environment can have missing dependencies which may prevent the exported workflow from building successfully.

We want to make sure that none of these issues are present when exporting the workflow to an image from the CLI for documentation purposes, so we introduce a default validation build step.

Fixes #1892


glopesdev:
@bruno-f-cruz mentioned these will break many self-contained "snippet" workflows (e.g. depending on subjects), since they are "valid" (i.e. no type is missing) but they will not build, since the subject declaration is purposefully missing.

PR #2279, opened 2025-05-07
Add CallBuilder operator

glopesdev:
To improve type manipulation and the flexibility of interfacing with existing modules we provide a new operator `Call` which is an adapter to construct a sequence selector based on calling an instance method over elements of the sequence.

This operator receives a single input sequence, and calls the appropriate instance method depending on the following properties:
- `MethodName`: the name of the instance method to call
- `InstanceSelector`: the name of a property on the type of the sequence elements which will be used as the instance on which to call the method
- `ArgumentSelector`: a set of zero or more properties contained in the type of the sequence elements which will be used as the arguments to the method call

If multiple overloads are present, the best match is selected using the same overload resolution rules used for resolving operator overloads.


glopesdev:
Feedback from DCM is that this still feels like a poor man's version of expression scripting, and may be worth reimagining this feature by expanding the auto-completion capabilities of `ExpressionTransform` et al. which could be done by setting Scintilla.NET `AutoCActive` and configuring hooks appropriately.

PR #2304, opened 2025-05-15
Add issue template for language proposals

### LINKED ISSUE #1738: Create issue templates to provide common structure for language proposals
Similar to C# language design repository and the harp-tech specifications, it would be useful to start providing some structure to evolve proposals for the language, the editor or broader package ecosystem, e.g. new package proposals.

glopesdev:
As the language evolves we want to make sure we are both as inclusive as possible, and as structured as possible, so that we do not miss the opportunity to document details of new language features as they are being investigated.

Here we introduce a template for creating language proposals with the goal of helping our language discussion and documentation.

Fixes #1738

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Propose a language idea or ask a question
+    url: https://github.com/orgs/bonsai-rx/discussions
+    about: Starting with discussion is the way to create a new proposal.
--- a/.github/ISSUE_TEMPLATE/proposal-template.md
+++ b/.github/ISSUE_TEMPLATE/proposal-template.md
@@ -1,0 +1,48 @@
+---
+name: Create a language specification
+about: For proposals that have been invited by a team member
+title: 'Feature name'
+labels: proposal
+assignees: ''
+
+---
+
+<!--
+Hello, and thanks for your interest in contributing to the Bonsai language standard! If you haven't been invited by a team member to open an issue, please instead open a discussion marked [proposal] at https://github.com/orgs/bonsai-rx/discussions and we'll try to give you feedback on how to get to an issue-ready proposal.
+
+New specification proposals should aim to fully fill out this template, at least up to and including detailed design. The sections on drawbacks, alternatives and unresolved questions may be omitted from the initial proposal.
+-->
+* [x] Proposed
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Not Started
+
+## Summary
+
+<!-- One paragraph explanation of the feature. -->
+
+## Motivation
+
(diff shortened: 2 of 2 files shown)


PathogenDavid:
IMO this is premature over-structuring. I think issue templates (particularly lengthy ones) are actually *less* inclusive - and are in fact intentionally *exclusive*. While they do establish clear expectations for what needs to be communicated, they create a huge barrier to entry and convey a sense of "Only those who are worthy can post things here." (This one doesn't even convey a sense of it, it outright says that.)

The templates used by the C# LDM fit their needs well, but their needs are significantly different from ours, their community is significantly larger than ours, and their internal processes are **significantly** different from ours.

I'm all for refining our development processes, but I don't think this is the path to something successful or helpful.

---------

If you're dead set on this, IMO it's a mistake to redirect people to discussions for informal proposals. We should reserve them for questions and announcements. Informal proposals will be drowned out otherwise.

We should also have clear messaging that submitting bug reports as issues is desired/acceptable.

We should use [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) instead of old-school markdown templates.

banchan86:
Agree with @PathogenDavid here, I think the template as is, is overly formal and restrictive, I would be hesitant to fill this out. 

I do support introducing a basic issue template that doesn't impose too high a barrier of entry. It might also nudge people to consider the issue more in depth. Something with just the headings (Description, Motivation, Possible Solution, Alternatives) and a basic question for each.  I recall seeing something done for either the harp-tech issues or discussions, but I can't find it now.

I also think we shouldn't direct people to discussions, I like keeping issues neat within their own repository.

glopesdev:
@PathogenDavid @banchan86 I'm closing this in favor of #2561, which merged a specification proposal template for the same goal. I merged #2561 without realizing the feedback here was still pending, which was my mistake, and I did not want it lost.

Re-reading it, your points stand, and the merged template did not address them. The specification template is currently the only issue template, so opening a new issue brings up the template chooser, whose single option is a long, formal proposal, with the blank issue link last. It also restricts the template to invited proposals and redirects informal ones to discussions, which both of you pushed back on.

I have opened #2586 to address this directly:

- A lightweight bug report form, ordered first, with a note that bug reports are always welcome.
- A lightweight feature request and informal proposal form using the Description, Motivation, Possible solution, and Alternatives structure @banchan86 suggested, so informal proposals have a home in issues rather than being sent to discussions.
- The specification template is kept for detailed design-stage specifications, but points early-stage ideas at the feature request form instead of the discussions.
- A config for the template chooser that keeps blank issues enabled and reserves discussions for questions and announcements.
- The two new templates are issue forms rather than markdown, as @PathogenDavid suggested, and they set a GitHub issue type while requiring only a description, to keep the friction minimal.

banchan86:
Saw the new templates, they look good to me, thank you for considering the feedback!

PR #2306, opened 2025-05-18
Update package template to introduce recommended repository structure

### LINKED ISSUE #1897: Consider updating package project template to follow release automation infrastructure
Following the changes to project infrastructure introduced in #1873 it might be a good idea to update the [package project template](https://github.com/bonsai-rx/bonsai/blob/8a63452c7de24352b203341c029a0d62ac575330/Bonsai.Templates/Bonsai.PackageTemplate/ProjectTemplate.csproj) to align with some of the modern patterns we adopted for .csproj files.

Specifically I was thinking of adopting artifacts output layout for all packages, but maybe there are some other ideas which could be easily ported over.

glopesdev:
The previous version of the package template was a C# project template where package metadata was interleaved with common build configurations. While useful for vanilla projects, this often led to confusion when used as a guidance for creating more complex multi-project packages.

Here we introduce two new packages:
  1. The `bonsaipackage` template is refactored into a solution template, and configures a complete repository structure containing reusable build infrastructure for either single- or multi-project package projects, laying a foundation for the recommended infrastructure for building package projects intended to be published in `nuget.org`.
  2. The `bonsailib` template is a simple C# class project with an example operator and a simple dependency to core which can be introduced to either a vanilla C# solution, or in a repository structure pre-configured by `bonsaipackage`. If the latter, all recommended package metadata will be reused and packaging will be immediately available.

Fixes #1897


glopesdev:
Feedback from DCM:
  - Investigate whether it is possible to enable child process debugging from within the `.csproj`.

PathogenDavid:
Haven't looked closely yet, but a few broad thoughts:

* As mentioned elsewhere, the Prefect reference template wasn't really designed with the community template in mind.
* I think updating the templates for 2.9 might be premature. I think we should dogfood things for a bit first.
* I think it'd also be better to release with some associated documentation, but I don't think we should/could hold up 2.9 for that to happen.
* One idea I wanted to explore with devclub is having two templates:
  * One template for John Neuroscientist who just wants to turn their `Extensions.csproj` into something reusable
  * One template for Bruno Meganerd who works is very serious about their packages and has needs similar to the Bonsai Foundation
  * (Migrating from one to another could be done with Prefect.)
* Could we move the templates to their own repository? I think the only thing actually keeping them here was the VSIX, and it was eliminated by this PR.

glopesdev:
> Could we move the templates to their own repository? I think the only thing actually keeping them here was the VSIX, and it was eliminated by this PR.

That is a good idea, you're totally right about the VSIX.

Doing that would also resolve my worry about not updating the templates for this release as it will effectively decouple the release cycles.

I will prepare the move and then delete everything in this PR (VSIX and .NET templates).

PathogenDavid:
@glopesdev It might be better to hold this PR until after 2.9 and allow the VSIX and the old templates release one last time (with `<PackageType>` added to the `PackageTemplate`)

Yanking them now creates an awkward situation for the [Create a Package](https://bonsai-rx.org/docs/articles/create-package.html) documentation since the Visual Studio extension won't actually exist anymore.

We *could* add a notice saying that the documentation isn't updated for 2.9 yet and to use 2.8.4, but since `Bonsai.Templates` will start being versioned separately from Bonsai I think it'd be easier to just do one final release of the old templates (with 2.9 compatibility) and we can update the documentation when we publish the new templates (under something other than version 2.9.)

glopesdev:
Thanks for bringing this up, that does make sense, although my strict preference in that case would be to release the new dotnet templates in their current state and simply update the documentation to describe how you initialise a project with the new templates (I have confirmed they are seen inside Visual Studio).

Will investigate but if this sounds good I will undo the last commit, merge for now, and make an issue in this repo for migrating the templates and then another issue in the docs to update documentation.

glopesdev:
One last thing to fix is the weird color inversion that is applied to dotnet templates icons, need to check more carefully but may be related to this: https://stackoverflow.com/questions/37663993/preventing-icon-color-and-size-distortions-when-bundling-a-visual-studio-project#38263209

PR #2364, opened 2025-06-23
Fix tooltip for `New` workflow button on toolbar

banchan86:
Still has the old keyboard shortcut rather than the new one.


glopesdev (approved):
Good catch!

Issue #1879, opened 2024-07-04
Consider allowing externalize properties to access inner properties


glopesdev:
Operator properties can be themselves complex structured types and it can happen often that we need to actually write to an inner member of that type. Externalizing inner properties would allow for example setting only the `X` component of a `Point`.

One of the main issues with the straightforward assignment is that even in C# it is not possible to directly assign sub-properties or sub-fields of value types exposed through properties, i.e. the below is not valid:

```c#
class Operator
{
    public Point Location { get; set; }
}

Operator op = new();
op.Location.X = 5; // this is not allowed;
op.Location = new Point(5, op.Location.Y); // instead we would need to do this
```

Some solutions might also require changes to `PropertyMapping` and `InputMapping` for consistency.

PathogenDavid:
```csharp
op.Location = new Point(5, op.Location.Y);
```

Technically the closer equivalent escape hatch would be:

```csharp
var location = op.Location;
location.x = 5;
op.Location = locatiojn;
```

PR #2370, opened 2025-07-03
Modify PropertyMap operator to access nested members

aacuevas:
Modifies the PropertyMap operator so it can be used to set inner members instead of just top-level properties of nodes, addressing #1879.

The Expression Builder creates a recursive assignment until it reaches the target member. For each intermediate type, it either does immediate asignment for reference types, or does a copy-modify-copyback for value types. We discussed these workings on the Dev Club meeting.

The editor for selecting the target property was more challenging than I thought, and I would like some feedback on it, as I am not sure I went through the best approach.

The issues is as following: The `MemberSelectorEditor` creates a dialog of class `MemberSelectorEditorDialog` which, in turn, creates a `MemberSelectorEditorController` which creates the actual tree view from a `Type`

This works with little modification from a normal object, i.e., directly interfacing with a single operator object. It gets complicated in the following two scenarios:
1. When the PropertyMapper is connected to several nodes. The common behavior for NameConverters in similar situations (and the previous behavior of the mapper) is to allow selection only of those properties that are common to all successors, so a custom list must be made.
2. When the PropertyMapper is connected to an encapsulated workflow with externalized properties, as this is no longer a Type.

I decided against adding multiple cases to `MemberSelectorEditorController` which would have had to be propagated, seemed too muddy. What I did is create a `MemberCollection` class inheriting from `TypeDelegator` which allows me to return an arbitrary list of `PropertyInfo` or  `FieldInfo` objects when asked, so I could store the custom list and pass that object as  `Type`

This solves issue number 1, but not completely 2. Externalized properties use component model `PropertyDescriptor`, however the controller is heavily designed towards using reflection so they cannot be used directly. My solution for this, which I am not sure is the best option, is to create a class deriving from `MemberInfo` which includes the type and name of these properties and add some extra logic to `TypeVisitor` to get them from types that implement these custom info objects. Originally I tried just creating a `PropertyInfo` object from the `PropertyDesciptor` but it cannot be fully mapped and, since only name and type are required by the editor, decided to go towards a simpler object even if it required some extra logic than creating an incomplete and potentially dangerous object.

A perhaps better approach would be if there were a way to obtain the underlying real property that an externalized property represents, but I admit I am not familiar enough yet with the workings of the workflow graph to get it properly. Also, while `PropertyDescriptor` has a `ComponentType` that points to the underlying type exposing the original property, and reflection could be used to get its `ProeprtyInfo`, the issue arises when the externalized property has a custom name. When this happens, the original property name is lost, at least through the public interface. If it were possible to add a way for `ExternalizedPropertyDescriptor` to get the original name, we could just use reflection to get the `PropertyInfo`, which would make things easier.

In any case, I welcome any and all advice on any potential changes to the editor code, or any other code.

Also, my naming sense for classes is terrible, so any suggestions on name changes or XMLcomments changes will be appreciated.

### DIFF
--- a/Bonsai.Core/Expressions/ExpressionBuilder.cs
+++ b/Bonsai.Core/Expressions/ExpressionBuilder.cs
@@ -1054,7 +1054,7 @@ namespace Bonsai.Expressions
             return BuildPropertyMapping(source, instance, propertyName, string.Empty);
         }
 
-        internal static Expression BuildPropertyMapping(Expression source, ConstantExpression instance, string propertyName, string sourceSelector)
+        internal static Expression UnwrapNestedWorkflowProperty(Expression source, ConstantExpression instance, string propertyName, Func<Expression, ConstantExpression, string, Expression> propertyMappingBuilderDelegate)
         {
             var element = instance.Value;
             if (element is IWorkflowExpressionBuilder workflowBuilder && workflowBuilder.Workflow != null)
@@ -1080,13 +1080,20 @@ namespace Bonsai.Expressions
                 {
                     var successorElement = GetWorkflowElement(successor.Target.Value);
                     var successorInstance = Expression.Constant(successorElement);
-                    argument = BuildPropertyMapping(argument, successorInstance, inputBuilder.workflowProperty.Name, sourceSelector);
+                    argument = propertyMappingBuilderDelegate(argument, successorInstance, inputBuilder.workflowProperty.Name);
                 }
                 return argument;
             }
+            return null;
+        }
+
+        internal static Expression BuildPropertyMapping(Expression source, ConstantExpression instance, string propertyName, string sourceSelector)
+        {
+            var argument = UnwrapNestedWorkflowProperty(source, instance, propertyName, (source, instance, propertyName) => BuildPropertyMapping(source, instance, propertyName, sourceSelector));
+            if (argument != null) { return argument; }
 
             MemberExpression property = default;
-            if (element is ICustomTypeDescriptor typeDescriptor)
+            if (instance.Value is ICustomTypeDescriptor typeDescriptor)
             {
                 var propertyInfo = instance.Type.GetProperty(propertyName);
                 if (propertyInfo == null)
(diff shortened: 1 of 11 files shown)


banchan86:
Hi @aacuevas I was trying it out and it works well with for example the `Region of Interest` property in the `Crop` operator from the `Bonsai.Vision` package. However I noticed it does not work with for example `CaptureProperties` in the `VideoCapture` operator from the `Bonsai.Video` package. Is it because that property is a property collection? How hard would it be to extend it to that? Sorry if I missed something out during the previous discussions, and if this is out of scope for the PR, but wonderful work so far and I look forward to seeing this feature!

aacuevas:
Hi! Sorry, I just came back from vacation.

Maybe it is worth discussing, but I believe collections are out of the scope of this PR.

The way the `PropertyMap` works is by traversing the class hierarchy through reflection and building an assignment expression. This is done through information retrieved by the types, not requiring the instances themselves.
However, Collections are dynamic containers that do require access to the specific instance to know their contents.

So, my take on this, is that while an operator to manipulate collection properties could be indeed useful, it is out of the scope of the current nested `PropertyMap`, as they have significantly different requirements. I am open to suggestions, though.

aacuevas:
I was reviewing some things and there is something that lacks in this implementation: The editor for the target properties should filter out non-externalizable properties, same as the NameConverter for the original `PropertyMap` does. Right now, it just shows everything, which should not be the case.
In any case, since this editor implementation is the part that needs the most review, I will wait until that to add this filtering, in case it would be better to redo with a complete different approach.

In a related topic, I noticed that even with the current `PropertyMap` in Bonsai, even if the NameConverter filters the non-externalizable properties, they can still be input in the field by hand, and this will work. (A funny example, one could change the Name and Description properties of any kind of nested workflow or expression builder). I don't know if this should be filtered or is a case of "if a user is trying so hard to break something, just let them"

jonnew:
Since it seems like this repo is seeing some activity, I want to resurface this idea. its something that a lot of users ask for and I think would be a big usability benefit for the editor.

Issue #2431, opened 2025-09-26
Setting WindowState to Maximized does not work the first time


glopesdev:
This seems related to an issue with `VisualizerWindow`:

<img height="70" alt="Image" src="https://github.com/user-attachments/assets/32bf1370-aeff-4593-8f98-90c3b5538e93" />

Setting the `WindowState` to `Maximized` on a fresh instance of this workflow will not set that property until the workflow is run again (or an existing layout file exists which sets the property).

bruno-f-cruz:
Another interesting quirk of this issue, if you set `Size` the fullscreen suddenly works.

PR #2559, opened 2026-05-15
Restore visualizer window bounds before Load

### LINKED ISSUE #2558: Late resize during `TypeVisualizerWindow.Load` can crash visualizers with native rendering contexts
When a visualizer window is reopened with a previously saved layout, the editor resizes the window inside the `Form.Load` event handler, i.e. after the form handle has been created along with any child control handles, including OpenGL or ImGui contexts created by the visualizer. The late resize is visually inconsequential on classic GDI visualizers, but it appears to corrupt visualizers that allocate a native rendering context tied to the window handle, leading to hard-to-reproduce crashes such as the one reported in bonsai-rx/imgui#33.

### Where the resize comes from

Two paths adjust the size of a `TypeVisualizerWindow`:

1. Default-size path, in [`Bonsai.Design/TypeVisualizerWindow.cs#L21-L31`](https://github.com/bonsai-rx/bonsai/blob/848b4298b7e67ad5d920f147a918b1dae40a6bd9/Bonsai.Design/TypeVisualizerWindow.cs#L21-L31). The `IDialogTypeVisualizerService.AddControl` method sets `ClientSize` to the size of the top-level control of the visualizer. This call normally fires from the visualizer-defined `Load` method, used across visualizers in `Bonsai.Design.Visualizers`, `Bonsai.Dsp.Design`, `Bonsai.Vision.Design`, etc. At this stage the form handle has not yet been created.

2. Restore-layout path, in [`Bonsai.Editor/Layout/WindowLauncher.cs#L35-L42`](https://github.com/bonsai-rx/bonsai/blob/848b4298b7e67ad5d920f147a918b1dae40a6bd9/Bonsai.Editor/Layout/WindowLauncher.cs#L35-L42). `WindowLauncher.Show` subscribes `VisualizerWindow.Load` and inside that handler assigns `LayoutBounds = bounds`, effectively setting `Form.Bounds`, to override path 1 with the persisted layout. By the time the `Load` event fires, the form handle exists, the child control has had its handle created, and any native context bound to that handle has been initialized at the default size.

The two paths interact: the form handle is created at the visualizer default size, then immediately resized to the persisted size.

### Why this is a problem

Some visualizers create native rendering surfaces when the control handle is first created. When the parent form is resized right after `Load`, the child control receives `OnSizeChanged` and `OnResize` notifications. Depending on the GL or ImGui binding, this can recreate or rebind the GL context against new dimensions, trigger framebuffer reallocation in the middle of the initial control setup, or invalidate state the visualizer assumed was stable after `Load`.

We have not pinned down the exact failure mode inside the affected backend, and the crash does not reproduce reliably; it depends on window manager timing, DPI, the size delta, and possibly the driver. Empirically, ensuring the window opens at its final size and is not resized at any point during initialization eliminates the crash across all the configurations we tested.

### Why it took so long to spot

On managed visualizers the late resize is invisible, since a single repaint absorbs it; no one had reason to investigate the ordering before.

The crash only manifests when three conditions hold together: a saved layout exists, the saved size differs from the visualizer default size, and the visualizer uses a native rendering context that is sensitive to early resize. Many users never hit all three.

The relevant code reads naturally; "restore the layout on Load" is a familiar pattern in WinForms, so the ordering hazard is not obvious from inspection.

### Open design questions

Should `IDialogTypeVisualizerService.AddControl` keep its side effect of sizing the window at all? Conceptually, a service for adding controls and a policy for sizing the form are two different things, and conflating them is what makes the default-size path overlap with the restore-layout path.

Should `TypeVisualizerWindow.AddControl` be made `virtual` so consumers can override the policy without explicit interface re-implementation?

Are there other late-init code paths, such as the observer subscription in `VisualizerWindowLauncher.InitializeComponents`, that could be hardened against similar timing assumptions?

### Verification

Affected: bonsai-rx/imgui#33, ImGui visualizers with a saved layout at non-default bounds.

Tested: extensive runs with the fix show no crash; visually identical behavior on classic visualizers.

glopesdev:
Move layout restore from the `Form.Load` event handler into the logic of the launcher `Show` method, before the window handle is created. Avoids resizing the control after handle creation, which can crash visualizers with native rendering contexts.

`TypeVisualizerWindow.AddControl` was marked as virtual so we can override its behavior in the launcher to preserve restored bounds against the default auto-size side effect.

Fixes #2558

### DIFF
--- a/Bonsai.Design/TypeVisualizerWindow.cs
+++ b/Bonsai.Design/TypeVisualizerWindow.cs
@@ -18,7 +18,7 @@ namespace Bonsai.Design
         }
 
         /// <inheritdoc/>
-        public void AddControl(Control control)
+        public virtual void AddControl(Control control)
         {
             ClientSize = control.Size;
             if (control.MinimumSize != Size.Empty)
--- a/Bonsai.Editor/Layout/WindowLauncher.cs
+++ b/Bonsai.Editor/Layout/WindowLauncher.cs
@@ -32,15 +32,13 @@ namespace Bonsai.Design
             if (VisualizerWindow == null)
             {
                 VisualizerWindow = CreateVisualizerWindow(provider);
-                VisualizerWindow.Load += delegate
+
+                var bounds = Bounds;
+                if (!bounds.IsEmpty && (SystemInformation.VirtualScreen.IntersectsWith(bounds) || WindowState != FormWindowState.Normal))
                 {
-                    var bounds = Bounds;
-                    if (!bounds.IsEmpty && (SystemInformation.VirtualScreen.IntersectsWith(bounds) || WindowState != FormWindowState.Normal))
-                    {
-                        VisualizerWindow.LayoutBounds = bounds;
-                        VisualizerWindow.WindowState = WindowState;
-                    }
-                };
+                    VisualizerWindow.LayoutBounds = bounds;
+                    VisualizerWindow.WindowState = WindowState;
+                }
 
                 VisualizerWindow.FormClosed += delegate
                 {
@@ -82,6 +80,7 @@ namespace Bonsai.Design
 
         protected class LauncherWindow : TypeVisualizerWindow
         {
+            bool hasRestoredBounds;
             SizeF inverseScaleFactor;
             SizeF scaleFactor;
 
@@ -98,8 +97,18 @@ namespace Bonsai.Design
                     if (layoutBounds.Width > 0)
                     {
                         Bounds = layoutBounds;
+                        hasRestoredBounds = true;
                     }
                 }
+            }
+
+            public override void AddControl(Control control)
+            {
+                // Prefer restored bounds over base auto-size
+                var restoredBounds = Bounds;
+                base.AddControl(control);
+                if (hasRestoredBounds)
+                    Bounds = restoredBounds;
             }
 
             protected override void ScaleControl(SizeF factor, BoundsSpecified specified)


bruno-f-cruz:
After a few dozen trials, I am still getting the same error. I will leave a MRE here just in case:

```xml
<?xml version="1.0" encoding="utf-8"?>
<WorkflowBuilder Version="2.9.1-refs-pull-2559-merge-ci495+6a63d31f7323890d77ad610e79798764ded4392f"
                 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                 xmlns:rx="clr-namespace:Bonsai.Reactive;assembly=Bonsai.Core"
                 xmlns:imvizui="clr-namespace:Bonsai.ImGui.Visualizers;assembly=Bonsai.ImGui.Visualizers.Design"
                 xmlns:implot="clr-namespace:Bonsai.ImPlot;assembly=Bonsai.ImPlot"
                 xmlns:wie="clr-namespace:Bonsai.Windows.Input;assembly=Bonsai.Windows.Input"
                 xmlns:imviz="clr-namespace:Bonsai.ImGui.Visualizers;assembly=Bonsai.ImGui.Visualizers"
                 xmlns:gui="clr-namespace:Bonsai.Gui;assembly=Bonsai.Gui"
                 xmlns:ui="clr-namespace:Bonsai.Design;assembly=Bonsai.Design"
                 xmlns="https://bonsai-rx.org/2018/workflow">
  <Workflow>
    <Nodes>
      <Expression xsi:type="rx:Defer">
        <Name>Ethogram</Name>
        <Workflow>
          <Nodes>
            <Expression xsi:type="imvizui:ImPlotVisualizerBuilder" />
            <Expression xsi:type="rx:PublishSubject">
              <Name>Tick</Name>
            </Expression>
            <Expression xsi:type="WorkflowOutput" />
            <Expression xsi:type="SubscribeSubject">
              <Name>Tick</Name>
            </Expression>
            <Expression xsi:type="implot:PlotBuilder">
              <implot:Visible>true</implot:Visible>
              <implot:Size>
                <implot:X>-1</implot:X>
                <implot:Y>0</implot:Y>
              </implot:Size>
              <implot:Flags>None</implot:Flags>
              <implot:FlagsX>AutoFit</implot:FlagsX>
              <implot:FlagsY>AutoFit</implot:FlagsY>
            </Expression>
            <Expression xsi:type="rx:PublishSubject">
              <Name>Plot</Name>
            </Expression>
            <Expression xsi:type="Combinator">
              <Combinator xsi:type="wie:MouseMove" />
            </Expression>
            <Expression xsi:type="imviz:RollingBufferBuilder">
              <imviz:Capacity>1000</imviz:Capacity>
              <imviz:ValueSelector>X,Y</imviz:ValueSelector>
            </Expression>
            <Expression xsi:type="imviz:LatestOnNewFrame">
              <Name>Plot</Name>
            </Expression>
            <Expression xsi:type="imviz:PlotLineSeriesBuilder">
              <imviz:Visible>true</imviz:Visible>
              <imviz:Flags>None</imviz:Flags>
            </Expression>
          </Nodes>
          <Edges>
            <Edge From="0" To="1" Label="Source1" />
            <Edge From="1" To="2" Label="Source1" />
            <Edge From="3" To="4" Label="Source1" />
            <Edge From="4" To="5" Label="Source1" />
            <Edge From="6" To="7" Label="Source1" />
            <Edge From="7" To="8" Label="Source1" />
            <Edge From="8" To="9" Label="Source1" />
          </Edges>
        </Workflow>
      </Expression>
      <Expression xsi:type="VisualizerMapping" />
      <Expression xsi:type="rx:Defer">
        <Name>PatchState</Name>
        <Workflow>
          <Nodes>
            <Expression xsi:type="imvizui:ImPlotVisualizerBuilder" />
            <Expression xsi:type="rx:PublishSubject">
              <Name>Tick</Name>
            </Expression>
            <Expression xsi:type="WorkflowOutput" />
            <Expression xsi:type="SubscribeSubject">
              <Name>Tick</Name>
            </Expression>
            <Expression xsi:type="implot:PlotBuilder">
              <implot:Visible>true</implot:Visible>
              <implot:Size>
                <implot:X>-1</implot:X>
                <implot:Y>0</implot:Y>
              </implot:Size>
              <implot:Flags>None</implot:Flags>
              <implot:FlagsX>AutoFit</implot:FlagsX>
              <implot:FlagsY>AutoFit</implot:FlagsY>
            </Expression>
            <Expression xsi:type="rx:PublishSubject">
              <Name>Plot</Name>
            </Expression>
            <Expression xsi:type="Combinator">
              <Combinator xsi:type="wie:MouseMove" />
            </Expression>
            <Expression xsi:type="imviz:RollingBufferBuilder">
              <imviz:Capacity>1000</imviz:Capacity>
              <imviz:ValueSelector>X,Y</imviz:ValueSelector>
            </Expression>
            <Expression xsi:type="imviz:LatestOnNewFrame">
              <Name>Plot</Name>
            </Expression>
            <Expression xsi:type="imviz:PlotLineSeriesBuilder">
              <imviz:Visible>true</imviz:Visible>
              <imviz:Flags>None</imviz:Flags>
            </Expression>
          </Nodes>
          <Edges>
            <Edge From="0" To="1" Label="Source1" />
            <Edge From="1" To="2" Label="Source1" />
            <Edge From="3" To="4" Label="Source1" />
            <Edge From="4" To="5" Label="Source1" />
            <Edge From="6" To="7" Label="Source1" />
            <Edge From="7" To="8" Label="Source1" />
            <Edge From="8" To="9" Label="Source1" />
          </Edges>
        </Workflow>
      </Expression>
      <Expression xsi:type="VisualizerMapping" />
      <Expression xsi:type="gui:TableLayoutPanelBuilder">
        <gui:Name>DataVisualization</gui:Name>
        <gui:Enabled>true</gui:Enabled>
        <gui:Visible>true</gui:Visible>
        <gui:ColumnCount>1</gui:ColumnCount>
        <gui:RowCount>3</gui:RowCount>
        <gui:ColumnStyles />
        <gui:RowStyles />
        <gui:CellSpans />
      </Expression>
      <Expression xsi:type="ui:VisualizerWindow">
        <ui:Text>VrForaging</ui:Text>
        <ui:Visible>true</ui:Visible>
        <ui:Location xsi:nil="true" />
        <ui:Size>
          <ui:Width>800</ui:Width>
          <ui:Height>800</ui:Height>
        </ui:Size>
        <ui:WindowState xsi:nil="true" />
      </Expression>
    </Nodes>
    <Edges>
      <Edge From="0" To="1" Label="Source1" />
      <Edge From="1" To="4" Label="Source1" />
      <Edge From="2" To="3" Label="Source1" />
      <Edge From="3" To="4" Label="Source2" />
      <Edge From="4" To="5" Label="Source1" />
    </Edges>
  </Workflow>
</WorkflowBuilder>
```

glopesdev:
Closing this. The crash it was opened for, #2558 and bonsai-rx/imgui#33, was traced to a `WndProc` context-routing bug on the imgui side and fixed there on the wndproc-context branch. Instrumentation traces from the affected machine show the launcher path completes cleanly even on crashing runs, so the late resize during `Load` is not the root cause.

That leaves this as a cleanup rather than a fix. Restoring the layout before the window handle is created and overriding `AddControl` so the restored bounds survive the default auto-size is reasonable on its own, since a resize between handle creation and `Load` can cause issues for any future visualizer that allocates native resources early. However, it is not needed for any current bug, so I would rather revisit it as a deliberate cleanup if the need comes up again.

The separate maximize-on-first-run bug, #2431, is handled by #2583.

